PR #10, opened 2023-10-20
Blowout user schema; add sessions

matoszz:
- Added some relevant user fields
- found out you could use "mix-ins" [ref](https://entgo.io/docs/schema-mixin/) to control fields you'd always want in your schemas for stuff like created at, updated at, etc., added the audit mixin and then updated the existing schemas with it
- noted a few "TO-DO"'s to determine what model you need to use for `Byte` as many things like a password hash / key you shouldn't store or pass as a string
- added more relevant comments to the files and also appease the linter


internal/ent/schema/user.go
```diff
@@ -21,26 +22,108 @@ type User struct {
...
+		field.String("avatar_local_file").
+			Comment("The user's local avatar file").
+			MaxLen(255).
+			Optional().
+			Nillable(),
+		field.Time("avatar_updated_at").
+			Comment("The time the user's (local) avatar was last updated").
+			Optional().
+			Nillable(),
+		field.Time("silenced_at").
```
sfunkhouser, line 69: what does silenced mean here?

internal/ent/schema/user.go
```diff
@@ -21,26 +22,108 @@ type User struct {
...
 				return err
 			}),
 		field.UUID("id", uuid.UUID{}).Default(uuid.New).Unique(),
-		field.Time("created_at").
-			Default(time.Now).
-			Immutable().
-			Annotations(&entsql.Annotation{
-				Default: "CURRENT_TIMESTAMP",
-			}),
+		field.String("display_name").
```
sfunkhouser, line 33: Should we also have first / last name?

internal/ent/schema/user.go
```diff
@@ -21,26 +22,108 @@ type User struct {
...
+		field.Time("avatar_updated_at").
+			Comment("The time the user's (local) avatar was last updated").
+			Optional().
+			Nillable(),
+		field.Time("silenced_at").
+			Comment("The time the user was silenced").
+			Optional().
+			Nillable(),
+		field.Time("suspended_at").
+			Comment("The time the user was silenced").
```
golanglemonade, line 74:
```suggestion
			Comment("The time the user was suspended").
```

internal/ent/schema/user.go
golanglemonade, line 33: do we also want a first/last name?

internal/ent/schema/user.go
golanglemonade, line 69: what does silenced mean here?

internal/ent/schema/user.go
matoszz: yea - adding

internal/ent/schema/user.go
matoszz: I was thinking it would be useful to have w/r/t notification channels + activity feeds

golanglemonade (approved):

PR #17, opened 2023-10-23
Atlas - use atlas config instead of goose

golanglemonade:
This PR fixes our atlas CI to use atlas native migrations instead of goose. This allows us to use atlas-ci linter: 

https://datum.atlasgo.cloud/dirs/4294967296


matoszz (approved):
GOOSE UP!

PR #21, opened 2023-10-26
Groups

matoszz:
- Adds group and group settings to the schema + API
- updates README to point back to the go-template so we're writing in one place
- adds excludes to the gitignore


internal/ent/schema/group.go
```diff
@@ -0,0 +1,54 @@
...
+// Group holds the schema definition for the Group entity.
+type Group struct {
+	ent.Schema
+}
+
+// Fields of the Group.
+func (Group) Fields() []ent.Field {
+	return []ent.Field{
+		field.UUID("id", uuid.UUID{}).Default(uuid.New).Unique(),
+		field.String("name").NotEmpty().Unique(),
```
golanglemonade, line 22:
```suggestion
		field.String("name").NotEmpty(),
```

This shouldn't be unique across all organizations

golanglemonade (approved):

PR #26, opened 2023-10-28
Add privacy

matoszz:
I created a new directory Docker with a copy of the existing dockerfile and plan to uplift the Compose setup in a different PR


internal/ent/schema/group.go
```diff
@@ -15,10 +16,18 @@ type Group struct {
 	ent.Schema
 }
 
+// Mixin of the User
```
golanglemonade, line 19:
```suggestion
// Mixin of the Group
```

Taskfile.yaml
```diff
@@ -0,0 +1,51 @@
...
+tasks:
+  clean:
+    cmds:
+      - go run ./internal/entclean/main.go --path=internal/ent {{ .CLI_ARGS }}
+      - "rm -rf internal/ent/generated/^doc.go"
+      - "rm -rf internal/api/^resolver.go"
+      - "rm -f schema/ent.graphql"
+      - rm -f schema.graphql
+      - rm -rf internal/testclient/
+  gen:
```
golanglemonade, line 12: Can this be ent, since thats what is being generated ?

```suggestion
  ent:
```

Taskfile.yaml
```diff
@@ -0,0 +1,51 @@
...
+    - go generate ./...
+  install:
+    cmds:
+      - go install github.com/go-task/task/v3/cmd/task@latest
+      - go install entgo.io/ent/cmd/ent@latest
+      - curl -sSf https://atlasgo.sh | sh
+      - curl -sSL https://rover.apollo.dev/nix/latest | sh
+      - defer: { task: go:tidy }
+  gqlgen:
```
golanglemonade, line 21:
```suggestion
  graph:
    cmds:
        - bash scripts/gen_graphql.sh
  gqlgen:
```

docker/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM golang:1.21 as builder
```
golanglemonade, line 1: You added this here but I don't see a remove of the `Dockerfile` in the root directory

.golangci.yaml
```diff
@@ -4,6 +4,9 @@ linters-settings:
   gofumpt:
     extra-rules: true
 
+skip-dirs:
+  - internal/ent/entclean
```
golanglemonade, line 8:
```suggestion
  - internal/entclean
```

internal/ent/mixin/basepolicymixin.go
```diff
@@ -0,0 +1,29 @@
...
+}
+
+// Policy defines the privacy policy of the BaseMixin.
+func (BaseMixin) Policy() ent.Policy {
+	return privacy.Policy{
+		Query: privacy.QueryPolicy{
+			// Deny any operation in case there is no "viewer context".
+			rule.DenyIfNoViewer(),
+			// Allow admins to query any information.
+			rule.AllowIfAdmin(),
```
golanglemonade, line 22: I think you are missing the final privacy policy, either `privacy.AlwaysDenyRule()` or ` privacy.AlwaysAllowRule()`.

matoszz: should it be:
```
			rule.DenyIfNoViewer(),
			rule.AllowIfAdmin(),
			privacy.AlwaysDenyRule(),
```

Taskfile.yaml
```diff
@@ -44,11 +44,17 @@ tasks:
       - task: atlas:lint
       - task: atlas:migrate
   atlas:create:
-    - go run db/create_migrations.go {{ .CLI_ARGS }}
+    cmds:
+      - |
+        echo "Enter the name of the migration:"
```
golanglemonade, line 49: see now I can get behind this over Makefiles 💯

internal/ent/mixin/basepolicymixin.go
golanglemonade: for Mutation to start yeah, and for Query `privacy.AlwaysAllowRule()` (you probably don't need the admin rule on the Query either.) This will most likely get more complicated but I think that is a good starting point

internal/ent/mixin/basepolicymixin.go
```diff
@@ -0,0 +1,32 @@
...
```
golanglemonade, line 23:
```suggestion
			rule.AllowIfAdmin(),
			privacy.AlwaysAllowRule()
```

.golangci.yaml
```diff
@@ -4,6 +4,12 @@ linters-settings:
   gofumpt:
     extra-rules: true
 
+run:
+  skip-dirs:
+    - internal/entclean
+  skip-files:
+    - internal/entclean/main.go
```
golanglemonade, line 11: why do you need both the file and the dir?

db/migrations/20231029062211_taskfile.sql
```diff
@@ -0,0 +1,2 @@
+-- Create "user_groups" table
```
golanglemonade, line 1: Why is the name of this migration taskfile?

matoszz: cuz it was generated when i was testing the taskfile ¯\_(ツ)_/¯

.golangci.yaml
matoszz: cuz the skip-dirs directive wasn't working but i think it's because i didn't have it nested under `run:`

db/migrations/20231029062211_taskfile.sql
golanglemonade: I mean, thats not super helpful for someone looking back at the migrations, its there so a human can easily find migrations/changes

golanglemonade (approved):

golanglemonade (approved):

PR #32, opened 2023-10-29
Adds echo-jwt middleware

golanglemonade:
- Adds the echo-jwt middleware
- Adds actor sub to echo context to be used by api resolvers and permissions checks


matoszz (approved):
we'll want to add / implement https://echo.labstack.com/docs/middleware/csrf + https://echo.labstack.com/docs/middleware/cors + https://echo.labstack.com/docs/middleware/secure and blow out the config options / helpers but this is a good start 👍

PR #34, opened 2023-10-29
Changes created_by and updated_by to UUIDs, adds custom scaler

golanglemonade:


internal/ent/mixin/audit_mixin.go
```diff
@@ -23,9 +24,9 @@ func (AuditMixin) Fields() []ent.Field {
 		field.Time("updated_at").
 			Default(time.Now).
 			UpdateDefault(time.Now),
-		field.Int("created_by").
+		field.UUID("created_by", uuid.UUID{}).
```
matoszz, line 27: this is supposed to map to a user?

golanglemonade: yeah

matoszz (approved):

PR #41, opened 2023-10-31
Remove memberships; make organization hierarchal

matoszz:
- Removes membership schema, associated edges, resolvers, etc.
- re-builds edges in between user, group, and organization
- adds name to integration
- creates an index on group which is what drives uniqueness of the group names within an organization
- Adds `parent_organization_id` field to Organization; creates `edge.To children` to make the Organization object a recursive O2M relation with itself (organizations can have organization; each "child" organization has only 1 parent


internal/ent/schema/organization.go
```diff
@@ -32,15 +32,38 @@ func (Organization) Fields() []ent.Field {
...
 				entgql.OrderField("name"),
+				entgql.Skip(entgql.SkipWhereInput),
+			),
+		field.String("description").
+			Comment("An optional description of the Organization").
+			Optional().
+			Annotations(
+				entgql.Skip(entgql.SkipWhereInput),
+			),
+		field.UUID("parent_organization_id", uuid.UUID{}).Default(uuid.New).Unique().Optional().Immutable().
```
golanglemonade, line 43: This shoudn't have a default value, and should be nullable for a `root` org. I don't think it needs to be unique either, a root org could have multiple child orgs. 
```suggestion
		field.UUID("parent_organization_id", uuid.UUID{}).
```

internal/ent/schema/integration.go
```diff
@@ -35,7 +40,8 @@ func (Integration) Fields() []ent.Field {
 // Edges of the Integration
 func (Integration) Edges() []ent.Edge {
 	return []ent.Edge{
-		edge.From("organization", Organization.Type).Ref("integrations").Unique().Required(),
+		//		edge.From("organization", Organization.Type).Ref("integrations").Unique().Required(),
```
golanglemonade, line 43: Delete this? 
```suggestion
```

internal/ent/schema/organization.go
```diff
@@ -32,15 +32,38 @@ func (Organization) Fields() []ent.Field {
...
+			Comment("An optional description of the Organization").
+			Optional().
+			Annotations(
+				entgql.Skip(entgql.SkipWhereInput),
+			),
+		field.UUID("parent_organization_id", uuid.UUID{}).
+			Comment("The ID of the parent organization for the organization.").
+			Annotations(
+				entgql.Type("ID"),
+				entgql.Skip(entgql.SkipWhereInput, entgql.SkipMutationUpdateInput, entgql.SkipType),
```
golanglemonade, line 47: Any reason why we wouldn't want to query organization based on `parent_organization_id`? 
```suggestion
				entgql.Skip(entgql.SkipMutationUpdateInput, entgql.SkipType),
```

golanglemonade (approved):

PR #46, opened 2023-11-03
Adding TLS Config 

golanglemonade:
Provides both an http or https server using the `--http` flag. If not passed, it will default to an http server (current state). If `--http` is passed, it will use `:443` as the default port and require `https:` requests. 

With the basic `task run-dev`, http server is provided: 

```
curl -s -H "Content-Type: application/json" -H "Authorization: Bearer $JWT_TEST" -X POST http://localhost:17608/query   --data-raw $'{"query":"query Organization($organizationId: ID\u0021) {\\n  organization(id: $organizationId) {\\n    id\\n    name\\n  }\\n}","variables":{"organizationId":"730cf793-0881-4a6c-b3ff-1e8dd1fa0370"},"operationName":"Organization"}'   |jq
{
  "data": {
    "organization": {
      "id": "730cf793-0881-4a6c-b3ff-1e8dd1fa0370",
      "name": "datum-test-14"
    }
  }
}
```

Testing with a self-signed cert: 
```
go run main.go serve  --debug --pretty --dev --oidc=false --https --ssl-cert=server.crt --ssl-key=server.key
```

```
curl -k -H "Content-Type: application/json" -H "Authorization: Bearer $JWT_TEST" -X POST https://localhost:17608/query   --data-raw $'{"query":"query Organization($organizationId: ID\u0021) {\\n  organization(id: $organizationId) {\\n    id\\n    name\\n  }\\n}","variables":{"organizationId":"730cf793-0881-4a6c-b3ff-1e8dd1fa0370"},"operationName":"Organization"}'   |jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   314  100    94  100   220   2164   5064 --:--:-- --:--:-- --:--:--  8263
{
  "data": {
    "organization": {
      "id": "730cf793-0881-4a6c-b3ff-1e8dd1fa0370",
      "name": "datum-test-14"
    }
  }
}
```

There is an `auto-cert` option with letsencrypt, this cannot provide certificates for localhost.

### DIFF
--- a/internal/echox/config.go
+++ b/internal/echox/config.go
@@ -203,3 +203,11 @@ func (c Config) WithTLSConfig() Config {
 
 	return c
 }
+
+// WithTLSCerts sets the TLS Cert and Key locations
+func (c Config) WithTLSCerts(certFile, certKey string) Config {
+	c.TLSConfig.CertFile = certFile
+	c.TLSConfig.CertKey = certKey
+
+	return c
+}
--- a/internal/echox/config.go
+++ b/internal/echox/config.go
@@ -1,0 +1,205 @@
+package echox
+
+import (
+	"crypto/tls"
+	"time"
+
+	"github.com/labstack/echo/v4"
+)
+
+var (
+	// DefaultShutdownGracePeriod sets the default for how long we give the sever
+	// to shutdown before forcefully stopping the server.
+	DefaultShutdownGracePeriod = 5 * time.Second
+	// DefaultReadTimeout sets the default maximum duration for reading the entire request including the body.
+	DefaultReadTimeout = 15 * time.Second
+	// DefaultWriteTimeout sets the default maximum duration before timing out writes of the response.
+	DefaultWriteTimeout = 15 * time.Second
+	// DefaultIdleTimeout sets the default maximum amount of time to wait for the next request when keep-alives are enabled.
+	DefaultIdleTimeout = 30 * time.Second
+	// DefaultReadHeaderTimeout sets the default amount of time allowed to read request headers.
+	DefaultReadHeaderTimeout = 2 * time.Second
+	// DefaultCertFile is the default cert file location
+	DefaultCertFile = "server.crt"
+	// DefaultKeyFile is the default key file location
+	DefaultKeyFile = "server.key"
+)
+
+// Config is used to configure a new echo server
+type Config struct {
+	// Debug enables echo's Debug option.
+	Debug bool
+
+	// Dev enables echo's dev mode options.
+	Dev bool
+
+	// Listen sets the listen address to serve the echo server on.
+	Listen string
+
+	// HTTPS configures an https server
+	HTTPS bool
+
+	// ShutdownGracePeriod sets the grace period for in flight requests before shutting down.
+	ShutdownGracePeriod time.Duration
+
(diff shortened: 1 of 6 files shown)


internal/echox/config.go
matoszz, line 1: omg thank you for pulling all this in !

internal/echox/config.go
```diff
@@ -0,0 +1,235 @@
...
+	// DefaultIdleTimeout sets the default maximum amount of time to wait for the next request when keep-alives are enabled.
+	DefaultIdleTimeout = 30 * time.Second
+	// DefaultReadHeaderTimeout sets the default amount of time allowed to read request headers.
+	DefaultReadHeaderTimeout = 2 * time.Second
+	// DefaultCertFile is the default cert file location
+	DefaultCertFile = "server.crt"
+	// DefaultKeyFile is the default key file location
+	DefaultKeyFile = "server.key"
+	// DefaultTLSConfig is the default TLS config used when HTTPS is enabled
+	DefaultTLSConfig = &tls.Config{
```
matoszz, line 29: have you actually tested if this default config will work without a certificate or with a self signed?

matoszz, line 29: just so you know, this is about the most strict you can enforce (for a "perfect score" on evaluation tools) and is very likely not what we want / can have right out of the gate

internal/echox/server.go
```diff
@@ -0,0 +1,278 @@
...
+// or if the http serve fails.
+// A graceful shutdown will be attempted
+func (s *Server) ServeHTTPSWithContext(ctx context.Context, listener net.Listener) error {
+	logger := s.logger.With(zap.String("address", listener.Addr().String()))
+
+	logger.Info("starting https server")
+
+	srv := s.defaultServer()
+
+	srv.Handler = s.Handler()
```
matoszz, line 210: to use the autocert functionality I believe you also need `srv.AutoTLSManager.Cache = autocert.DirCache("/var/www/.cache")` and would also probably want `srv.Pre(middleware.HTTPSRedirect())`

internal/echox/config.go
golanglemonade: I tested with a self-signed cert (see PR description). It doesn't work without a certificate, but it doesn't use this if https isn't enabled.

internal/echox/server.go
golanglemonade: yeah, the cache is here: https://github.com/datumforge/datum/pull/46/files#diff-88e2bf36dc421317c9e4ff226971afb76cf02879754bb11c66c1115cc3d3192dR221

I'll add the middleware though

golanglemonade: AutoTLSManager Cache is set: https://github.com/datumforge/datum/pull/46/files#diff-88e2bf36dc421317c9e4ff226971afb76cf02879754bb11c66c1115cc3d3192dR223

golanglemonade: HTTPSRedirect will be a follow-up TODO, it requires running both the http and https server, which is out of scope for originally adding the TLS config

matoszz (approved):
we need to test the TLS configuration / let's encrypt stuff on a public IP / domain to ensure it functions but gotta get the code out there to do that :)

matoszz (approved):

PR #50, opened 2023-11-03
Create dependabot.yml

matoszz:
adds package ecosystem for gomod + github actions; will add docker and others at a later date


golanglemonade (approved):
My assumption is that this will do the same updates as renovate, I've never seen them used together but shouldn't hurt anything

PR #51, opened 2023-11-04
Use nanox.ID over UUID, but as a string

golanglemonade:


internal/ent/mixin/idmixin.go
```diff
@@ -0,0 +1,50 @@
...
+	"entgo.io/ent/schema/mixin"
+
+	"github.com/jaevor/go-nanoid"
+)
+
+const (
+	idLength = 21
+)
+
+var _ ent.Mixin = (*IDMixin)(nil)
```
matoszz, line 15: i don't think this is needed / not sure why it's here?

internal/ent/mixin/idmixin.go
```diff
@@ -0,0 +1,50 @@
...
+// IDMixin holds the schema definition for the ID
+type IDMixin struct {
+	mixin.Schema
+}
+
+// Fields of the IDMixin.
+func (IDMixin) Fields() []ent.Field {
+	return []ent.Field{
+		field.String("id").
+			Unique().
```
matoszz, line 26: I think we should remove this; everything I've seen says if you're using `Default` or `DefaultFunc` and want to ensure uniqueness you can create an index on the ID vs. specifying it on the field

matoszz (dismissed):

matoszz (approved):

PR #56, opened 2023-11-06
add gosec to linter

matoszz:


matoszz:
being addressed in another PR

PR #64, opened 2023-11-07
add refresh token

matoszz:
- adds package entx
- attempts to add support for JSON usage but oas still failing


internal/ent/schema/refreshtoken.go
```diff
@@ -0,0 +1,58 @@
...
+
+// RefreshToken holds the schema definition for the RefreshToken entity
+type RefreshToken struct {
+	ent.Schema
+}
+
+// Fields of the RefreshToken
+func (RefreshToken) Fields() []ent.Field {
+	return []ent.Field{
+		field.String("client_id").
```
golanglemonade, line 20: Why is this `field.String`, and everything else `field.Text`

db/migrations/atlas.sum
```diff
@@ -1,3 +1,5 @@
-h1:blUbVXx6kZZsjPCQlEtb7qP5HP3bySZe8di4hNyfprA=
```
golanglemonade, line 1: can you squash these two migrations into one?

matoszz: I planned to re-base all our migrations in another PR once the schemas are added if that's OK

PR #70, opened 2023-11-09
organization setting schema

matoszz:
- updates refreshtoken schema types to JSON as originally intended
- adds schema organization settings


internal/ent/schema/refreshtoken.go
```diff
@@ -19,8 +19,8 @@ func (RefreshToken) Fields() []ent.Field {
 	return []ent.Field{
 		field.Text("client_id").
 			NotEmpty(),
-		// TO DO figure out why JSON doesn't work with oas / ogent
-		field.Strings("scopes").Optional(),
+		field.JSON("scopes", []string{}).
```
golanglemonade, line 22: These are the same thing, any reason why you wouldn't use the built in `Strings`?

golanglemonade (approved):

PR #72, opened 2023-11-09
add personal access tokens

matoszz:
- adds the schema personal access tokens for user-generated tokens
- abilities field needs discussion / input to determine how we'd generate tokens with specific scopes


internal/ent/schema/personalaccesstoken.go
```diff
@@ -0,0 +1,67 @@
...
+type PersonalAccessToken struct {
+	ent.Schema
+}
+
+// Fields of the PersonalAccessToken.
+func (PersonalAccessToken) Fields() []ent.Field {
+	return []ent.Field{
+		field.String("name"),
+		field.String("user_id"),
+		field.String("token"),
```
golanglemonade, line 24: Should this be flagged as `Sensitive(),` https://entgo.io/docs/schema-fields/#sensitive-fields

golanglemonade (approved):

PR #76, opened 2023-11-10
Add test utils and organization crud resolver tests

golanglemonade:
Adds unit tests for the organization resolvers


internal/api/test.db
matoszz, line 1: should this be pushed or added to `.gitignore`?

internal/api/tools_test.go
```diff
@@ -0,0 +1,123 @@
...
+	ent "github.com/datumforge/datum/internal/ent/generated"
+	"github.com/datumforge/datum/internal/entdb"
+)
+
+var (
+	defaultDBURI = "file:ent?mode=memory&cache=shared&_fk=1"
+	EntClient    *ent.Client
+)
+
+func TestMain(m *testing.M) {
```
matoszz, line 29: does this get run with the tasks that are already in our taskfile?

internal/api/test.db
golanglemonade: whoops, no that was from testing, I'll remove.

golanglemonade: yeah, it will run with `task go:test`

internal/api/tools_test.go
golanglemonade: yeah it will run with `task go:test`

matoszz (approved):

PR #77, opened 2023-11-10
User settings

matoszz:
- adds user settings table and moves some attributes in there
- cleanup and validation of some additional fields


internal/ent/schema/user_settings.go
```diff
@@ -0,0 +1,84 @@
...
+		field.Enum("role").
+			NamedValues(
+				"User", "USER",
+				"Admin", "ADMIN",
+				"Owner", "OWNER",
+			).
+			Default("USER"),
+		field.JSON("permissions", []string{}).Default([]string{}),
+		// TO DO: not sure if this should be default true or false
+		field.Bool("email_confirmed").Default(true),
```
golanglemonade, line 62: I would say default to `false` and flipped to `true` after confirmation

golanglemonade (dismissed):

golanglemonade (approved):

PR #91, opened 2023-11-17
Adds a basic FGA model with organization, groups, subscriptions, and features

golanglemonade:
Model: 
image

- Adds tests cases for the model and the tests to the buildkite pipeline. 
- There are a lot of new commands added to the Taskfile, that aren't necessarily relevant to this specific update, but pulled from my main fga branch that didn't seem necessary to separate out.


internal/fga/testdata/tests.yaml
```diff
@@ -0,0 +1,331 @@
...
+    check: 
+      - user: organization:datum
+        object: subscription_tier:pro
+        assertions:
+          subscriber: true
+          subscriber_member: false
+      - user: organization:datum
+        object: subscription_tier:free
+        assertions:
+          subscriber: false
```
matoszz, line 197: shouldn't this be true of any tier?

PR #93, opened 2023-11-19
Adding authz with openfga

golanglemonade:
- Adds FGA library for client, tuple creation/deletion, and tuple checks 
- Allows all to continue to work without a JWT when `oidc` is disabled
- Uses ent `privacy` for all authz checks when getting resources
- Uses ent `Hooks` to create tuples during object update/creation/deletion
- Creates a mock client using gomock + make generate for calls made to FGA 
- Adds authz checks to organization get, create, update, delete (but not yet "get all organizations", this will be done in a follow-up PR)
- Adds interceptor to cli to accept `DATUM_ACCESS_TOKEN` from the envvars for authz checks 
- Currently removed `parent` from cli queries for organization, this will require more work on how we determine access in the hierarchical model, which will be done in a follow-up PR

To use the cli with auth, run the datum server using:

```
task run-dev-auth 
```

In your terminal, set the envvar:
```
export DATUM_ACCESS_TOKEN=$JWT_TEST   
```

Create an org using that JWT: 
```
go run cmd/cli/main.go org create -n "Lemonade Golang" -d "a new org created with the CLI"
{
  "createOrganization": {
    "organization": {
      "createdAt": "2023-11-24T13:41:16.662749-07:00",
      "createdBy": "",
      "displayName": "unknown",
      "id": "2h8fQ-XNuqi0oQftLTKBZ",
      "name": "Lemonade Golang",
      "updatedAt": "2023-11-24T13:41:16.662729-07:00",
      "updatedBy": ""
    }
  }
}
```

As that same user/same `JWT` you should be able to do a `GET`

```
 go run cmd/cli/main.go org get -i 2h8fQ-XNuqi0oQftLTKBZ
{
  "organization": {
    "createdAt": "2023-11-24T13:41:16.662749-07:00",
    "createdBy": "",
    "displayName": "unknown",
    "id": "2h8fQ-XNuqi0oQftLTKBZ",
    "name": "Lemonade Golang",
    "updatedAt": "2023-11-24T13:41:16.662729-07:00",
    "updatedBy": ""
  }
}
```

As another user: 
```
export DATUM_ACCESS_TOKEN=$JWT_TEST_ANOTHER
```

The `GET` should fail: 

```
go run cmd/cli/main.go org get -i 2h8fQ-XNuqi0oQftLTKBZ
Error: {"networkErrors":null,"graphqlErrors":[{"message":"you are not authorized to perform this action","path":["organization"]}]}
```


Taskfile.yaml
```diff
@@ -125,6 +123,10 @@ tasks:
     desc: opens the fga playground in a browser
     cmds: 
       - 'open "http://localhost:3000/playground"'
+  fga:model:transform:
+    desc: transforms the fga model to json
+    cmds:
+      - fga model transform --file=fga/model/datum.fga |jq  > fga/model/datum.json
```
matoszz, line 129: might be useful to leverage taskfile's ability to check for output files / checksum?

cmd/serve.go
```diff
@@ -136,7 +182,7 @@ func serve(ctx context.Context) error {
 	}
 
 	// add jwt middleware
-	if viper.GetBool("oidc.enabled") {
+	if oidcEnabled {
 		secretKey := viper.GetString("jtw.secretkey")
 		jwtConfig := createJwtMiddleware([]byte(secretKey))
```
matoszz, line 187: jtw vs. jwt?

internal/api/errors.go
```diff
@@ -5,4 +5,7 @@ import "errors"
 var (
 	// ErrInternalServerError is returned when an internal error occurs.
 	ErrInternalServerError = errors.New("internal server error")
+
+	// ErrNotFound is returned when a resource is not found or the user does not have permissions to the resource
```
matoszz, line 9: this should probably be 2 separate errors? resource not found vs. permissions to the resource are very different debugging paths

internal/api/organization.resolvers.go
```diff
@@ -36,6 +36,18 @@ func (r *mutationResolver) CreateOrganization(ctx context.Context, input generat
 		return nil, ErrInternalServerError
 	}
 
+	// // Add relationship tuples
+	// // When an organization is created, the user that created the org should have access
+	// // By default, that user should have an owner relationship
+	// // All other users assigned to the org will have a member relationship
+	// // Users can be promoted to admin of the organization
+	// if err = r.authz.CreateRelationshipTupleWithUser(ctx, ownerRelationship, fmt.Sprintf("%s:%s", objectType, org.ID)); err != nil {
+	// 	// TODO - rollback creation if permissions fail to be created
```
matoszz, line 45: what's the circumstance where it'd fail, but be created?

internal/fga/fga.go
```diff
@@ -0,0 +1,229 @@
...
+		}
+	}
+
+	// create model if ID was not configured
+	if config.ModelID == "" {
+		// create fga client with store ID
+		fgaClient, err := NewClient(
+			config.Host,
+			WithScheme(config.Scheme),
+			WithStoreID(config.StoreID),
```
matoszz, line 131: do we always want this to come from `config`?

internal/fga/fga.go
```diff
@@ -0,0 +1,229 @@
...
+	// Logger is the provided Logger
+	Logger *zap.SugaredLogger
+}
+
+type Config struct {
+	Name    string
+	Host    string
+	Scheme  string
+	StoreID string
+	ModelID string
```
matoszz, line 31: I could be missing it but where are these actually generated?

internal/api/organization.resolvers.go
golanglemonade: Well this code is being deleted, because I moved it took hooks rather than in the resolver. That being said, the answer is still relevant, which is if an organization gets created, but the permission fail to be created (e.g. issue connecting to the postgres database or something), we wouldn't want the org creation to say its successful but then the org to not be able to be accessed.

internal/fga/fga.go
golanglemonade: `StoreID` and `ModelID` can either be passed through to the server as flags/configs if one already exists. Otherwise,  `CreateFGAClientWithStore` creates a store and a model and sets the ids to be used later.

internal/api/errors.go
golanglemonade: Usually two separate logs, but most APIs I've seen don't want to give away permissions vs. not actually existing, so you will get (in REST land) a 404 error a lot of the times if you request an object and either it doesn't exist at all or you shouldn't have access to it.
This isn't actually used right now, so deleting it anyways.

aaguiarz:
Hi! I work on the OpenFGA team and found this PR :), I'm glad you are using it! Let me know if we can help in any way.

I wanted to mention that instead of storing the .json file, you can store an ".fga" file with the model itself, and use the FGA CLI to [write the model to OpenFGA](https://github.com/openfga/cli#write-authorization-model) when needed. You don't need to use the JSON format.

You can also use the CLI to run tests in the authorization model as part of CI/CD https://github.com/openfga/cli#run-tests-on-an-authorization-model

Hope it helps!

golanglemonade:
> Hi! I work on the OpenFGA team and found this PR :), I'm glad you are using it! Let me know if we can help in any way.
> 
> I wanted to mention that instead of storing the .json file, you can store an ".fga" file with the model itself, and use the FGA CLI to [write the model to OpenFGA](https://github.com/openfga/cli#write-authorization-model) when needed. You don't need to use the JSON format.
> 
> You can also use the CLI to run tests in the authorization model as part of CI/CD https://github.com/openfga/cli#run-tests-on-an-authorization-model
> 
> Hope it helps!

thanks @aaguiarz !

I was using the cli to transform from fga -> JSON but just added a [function](https://github.com/datumforge/datum/pull/93/files#diff-93e6506676af6067fb373e46279636e2ed41775948cf2ea9663643efe86bc0dcR241-R249) to do that instead.Thanks for pointing me towards that. 
 
I did come across the the cli to test the models, already added it to our CI ❤️, we were just discussing how much that helped us understand the model itself : https://github.com/datumforge/datum/blob/main/.buildkite/pipeline.yaml#L23-L28

Taskfile.yaml
golanglemonade: Removed this json model completely, there is now a function to convert FGA -> JSON when the model is created in the store.

matoszz:
hey @aaguiarz - I love that you reached out after seeing us working to implement FGA in our new stack! I'm a co-founder and CTO of an early-stage startup (also named Datum) and we're going to be tackling a broad range of capabilities in our software stack and plan to use openfga as a foundational component. I was curious if you had any thoughts / feedback around when (or if) it makes sense to have entirely separate authorization stores? We've read up on your docs regarding [running in production](https://openfga.dev/docs/getting-started/running-in-production) and use cases for passing the [authorization model ID](https://openfga.dev/docs/getting-started/immutable-models#potential-use-cases) which gave me the idea around having more segmented models vs. a monolithic one and being able to use the model ID's as a "air traffic controller" for which store should be called. 

Additionally we were curious about some guidance on what we should store in our own data store vs. what we should rely on the authorization data store for. Our general goal would be to reduce the number of databases we have to care and feed for in production and initially leaned towards re-storing information in our own data source but have since refactored and now are not storing user roles in our own database at all and aren't sure if this is the "right" method of approaching the problem with openfga. If a call / discussion is easier, happy to take the time to connect!

aaguiarz:
@golanglemonade Awesome! 

> I was using the cli to transform from fga -> JSON but just added a [function](https://github.com/datumforge/datum/pull/93/files#diff-93e6506676af6067fb373e46279636e2ed41775948cf2ea9663643efe86bc0dcR241-R249) to do that instead.Thanks for pointing me towards that.

What are you doing with the JSON file? You should not need it if you use the CLI to update the model in the store, you can just do `fga store write --file model.fga`

aaguiarz:
Nice meeting you @matoszz !

- In general, you would create another store if you'd use to solve independent authorization problems, e.g. permissions for your infrastructure vs permissions for your applications, or permissions for applications that have a different sets of users and types.
- The idea of the Model ID is to enable you roll out changes to production progressively, e.g. you update the model, you write additional tuples, and once you confirm everything is working fine you bump the version. You would not use it to have different conceptual models.
- In general, we don't recommend to use FGA as the source of truth for all type/relationships, as it might complicate some Search scenarios. E.g. if you want to know all the resources that a user can access, and there's a lot of them, relying on FGA to return it will not be feasible. Being able to write part of that query on your side will make it simpler. Simplifying how to address [this problem](https://openfga.dev/docs/interacting/search-with-permissions) is an area of investment for us, but it's hard to find a generic solution to all problems.

I'd love to chat at some moment, feel free to email me at andres dot aguiar at openfga.dev

generate.go
```diff
@@ -1,3 +1,5 @@
 package main
 
 //go:generate go run -mod=mod ./internal/ent/entc.go
+
+//go:generate mockgen -destination=internal/fga/mocks/client.go -package=mock_client --build_flags=--mod=mod github.com/openfga/go-sdk/client SdkClient,SdkClientListStoresRequestInterface,SdkClientCreateStoreRequestInterface,SdkClientGetStoreRequestInterface,SdkClientDeleteStoreRequestInterface,SdkClientReadAuthorizationModelsRequestInterface,SdkClientWriteAuthorizationModelRequestInterface,SdkClientReadAuthorizationModelRequestInterface,SdkClientReadLatestAuthorizationModelRequestInterface,SdkClientReadChangesRequestInterface,SdkClientReadRequestInterface,SdkClientWriteRequestInterface,SdkClientWriteTuplesRequestInterface,SdkClientDeleteTuplesRequestInterface,SdkClientCheckRequestInterface,SdkClientBatchCheckRequestInterface,SdkClientExpandRequestInterface,SdkClientListObjectsRequestInterface,SdkClientListRelationsRequestInterface,SdkClientReadAssertionsRequestInterface,SdkClientWriteAssertionsRequestInterface
```
matoszz, line 5: should we separate these out into different files / commands for sanity on the codegen process?

internal/api/errors.go
```diff
@@ -5,4 +5,7 @@ import "errors"
 var (
 	// ErrInternalServerError is returned when an internal error occurs.
 	ErrInternalServerError = errors.New("internal server error")
+
+	// ErrPermissionDenied is returned when the user is not authorized to perform the requested query or mutation
+	ErrPermissionDenied = errors.New("you are not authorized to perform this action")
```
matoszz, line 10: if it's possible / easy we should create separate errors for create, update, delete, etc., for future debugging

internal/api/tools_test.go
```diff
@@ -9,23 +9,33 @@ import (
...
 )
 
 var (
 	defaultDBURI = "file:ent?mode=memory&cache=shared&_fk=1"
 	EntClient    *ent.Client
 )
 
+const (
+	subClaim = "1234567890"
+	rawToken = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ.oGFhqfFFDi9sJMJ1U2dWJZNYEiUQBEtZRVuwKE7Uiak" //nolint:gosec
```
matoszz, line 36: interested to see if github flags this as a committed secret

internal/datumclient/auth.go
```diff
@@ -0,0 +1,24 @@
+package datumclient
```
matoszz, line 1: should we put this alongside the generated files which get written to this directory? In another branch to make a robust "clean script" I remove the generated client directory entirely which would also remove this file; we should ideally not couple our code with the generated code, or potentially use a header or something?

internal/ent/mixin/basepolicymixin.go
```diff
@@ -1,29 +1,10 @@
...
-	"github.com/datumforge/datum/internal/ent/privacy/rule"
 )
 
 // BaseMixin for all schemas in the graph.
 type BaseMixin struct {
 	mixin.Schema
 }
-
-// Policy defines the privacy policy of the BaseMixin.
-func (BaseMixin) Policy() ent.Policy {
```
matoszz, line 17: was this not updated with a different policy for some reason?

internal/fga/fga.go
```diff
@@ -0,0 +1,257 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	client.Ofga = fgaClient
+
+	return &client, err
+}
+
+// WithScheme sets the open fga scheme, defaults to "https"
```
matoszz, line 75: defaults to https but we configure for http - TLS is a "TODO" I assume?

matoszz (approved):
All comments are informational / feedback, and remaining work we'll write up in TO-DO's / issues. Great work @golanglemonade !

generate.go
golanglemonade: Doesn't really matter, its not code you'd probably ever read, but just pulls all the interfaces available in the client and creates a mock. I'd say leave it as is and we can break it up later if we see issues.

internal/api/errors.go
golanglemonade: Yep, very easy. Added an issue: https://github.com/datumforge/datum/issues/127

internal/api/tools_test.go
golanglemonade: it did, thats why I had to add `//nolint:gosec`

internal/datumclient/auth.go
golanglemonade: There is mixed generated code and not throughout, although I agree I don't like it. This doesn't need to be here though, so we can moved this to another package if we want, its just specific to creating the interceptor (in this case the authorization header) for the client.

internal/ent/mixin/basepolicymixin.go
golanglemonade: Chatted in slack, but for tracking sack, this was removed because it didn't work, and the new policies are more strict. We will come back to a default base policy: https://github.com/datumforge/datum/issues/106

internal/fga/fga.go
golanglemonade: yeah, when you run locally you need http because of localhost. This should easily work with `https`.

PR #96, opened 2023-11-20
fix case and pluralism mismatches

matoszz:
- squashes existing migrations back into `init`
- removes schemas (graphql and ent) with `_` in the file name
- updates generates files, clients, struct / function names from `*settings` -> `setting`
- fixes camel case naming in `/schema/`


golanglemonade (approved):

matoszz:
Verifying these commits were from me and my keysigning is currently disabled

PR #125, opened 2023-11-25
Add scaffolding for initial Prometheus metrics

### LINKED ISSUE #116: Add prometheus metrics
Given the maturity of the current stack we don't have a burning need for these metrics but should begin to add them to the core echo server and other components as a part of our production readiness. At a minimum we should have basic HTTP + graphQL metrics.

tylerauerbeck:
This adds the initial scaffolding for a `/metrics` endpoint which will return prometheus metrics. At the moment, no `datum` specific metrics have been added and we're just using the out of the box `echo` metrics. Whenever stories around useful metrics for `datum` are gathered, they can be added using `promauto` similar to the below:

Define your metric in an appropriate location:
```
var (
    // List of dynamic labels
    labelNames = []string{"example1", "example2"}

    // Create a counter collector
    exampleMetric = promauto.NewCounterVec(
        prometheus.CounterOpts{
            Name: "dummy_metric",
            Help: "Number of dummy_metrics",
        },
        labelNames,
    )
)
```

Then in an appropriate resolver/function, you can use them as necessary:
```
exampleMetric.WithLabelValues("my-metric", "something-value").Inc()
```

Resolves #116


matoszz:
@tylerauerbeck really appreciate your contributions <3

golanglemonade (approved):

PR #130, opened 2023-11-26
add passwd package

matoszz:
- adds passwd package and creates methods for using derived key functionality via argon2
- creates functions for referencing / checking stored passwords
- creates basic strength tests / checks


internal/passwd/errors.go
```diff
@@ -0,0 +1,24 @@
...
+	ErrUnableToVerify        = errors.New("cannot verify empty derived key or password")
+	ErrCannotParseDK         = errors.New("cannot parse encoded derived key, does not match regular expression")
+	ErrCannotParseEncodedEK  = errors.New("cannot parse encoded derived key, matched expression does not contain enough subgroups")
+	ErrDKProtocol            = fmt.Errorf("current code only works with the the dk protcol %q", dkAlg)
+	ErrExpectedDKVersion     = fmt.Errorf("expected %s version %d", dkAlg, argon2.Version)
+	ErrDKCouldNotParseMemory = fmt.Errorf("could not parse memory")
+	ErrDKParseTime           = fmt.Errorf("could not parse time")
+	ErrDKParseThreads        = fmt.Errorf("could not parse threads")
+	ErrDKParseSalt           = fmt.Errorf("could not parse salt")
+	ErrDKParseDK             = fmt.Errorf("could not parse derived key")
```
golanglemonade, line 23: This should all be `errors.New`, not `fmt.Errorf`, you are not formatting any variable

internal/passwd/errors.go
```diff
@@ -0,0 +1,24 @@
...
+
+// Error constants
+var (
+	ErrCannotCreateDK        = errors.New("cannot create derived key for empty password")
+	ErrCouldNotGenerate      = fmt.Errorf("could not generate %d length", dkSLen)
+	ErrUnableToVerify        = errors.New("cannot verify empty derived key or password")
+	ErrCannotParseDK         = errors.New("cannot parse encoded derived key, does not match regular expression")
+	ErrCannotParseEncodedEK  = errors.New("cannot parse encoded derived key, matched expression does not contain enough subgroups")
+	ErrDKProtocol            = fmt.Errorf("current code only works with the the dk protcol %q", dkAlg)
+	ErrExpectedDKVersion     = fmt.Errorf("expected %s version %d", dkAlg, argon2.Version)
```
golanglemonade, line 18: When formatting an error with parameters, its usually best to create a function. See example here: https://github.com/datumforge/datum/blob/main/internal/fga/errors.go#L25-L39

internal/passwd/errors.go
```diff
@@ -0,0 +1,24 @@
...
+import (
+	"errors"
+	"fmt"
+
+	"golang.org/x/crypto/argon2"
+)
+
+// Error constants
+var (
+	ErrCannotCreateDK        = errors.New("cannot create derived key for empty password")
```
golanglemonade, line 12: Comment all exported errors 🙏

internal/passwd/dk.go
```diff
@@ -0,0 +1,127 @@
...
+	dkKLen = uint32(32)        // the length of the derived key (32 bytes is the required key size for AES-256)
+)
+
+// Argon2 variables for the derived key (dk) algorithm
+var (
+	dkParse = regexp.MustCompile(`^\$(?P<alg>[\w\d]+)\$v=(?P<ver>\d+)\$m=(?P<mem>\d+),t=(?P<time>\d+),p=(?P<procs>\d+)\$(?P<salt>[\+\/\=a-zA-Z0-9]+)\$(?P<key>[\+\/\=a-zA-Z0-9]+)$`)
+)
+
+// CreateDerivedKey creates an encoded derived key with a random hash for the password.
+func CreateDerivedKey(password string) (_ string, err error) {
```
golanglemonade, line 35: Instead of having one named return in the function signature, I would just do: 

```suggestion
func CreateDerivedKey(password string) (string, error) {
```

And then line 41 will have to change to the following since `err` is not yet defined

```
if _, err := rand.Read(salt); err != nil {
```

internal/passwd/dk.go
```diff
@@ -0,0 +1,127 @@
...
+
+	dk := argon2.IDKey([]byte(password), salt, dkTime, dkMem, dkProc, dkKLen)
+	b64salt := base64.StdEncoding.EncodeToString(salt)
+	b64dk := base64.StdEncoding.EncodeToString(dk)
+
+	return fmt.Sprintf("$%s$v=%d$m=%d,t=%d,p=%d$%s$%s", dkAlg, argon2.Version, dkMem, dkTime, dkProc, b64salt, b64dk), nil
+}
+
+// VerifyDerivedKey checks that the submitted password matches the derived key.
+func VerifyDerivedKey(dk, password string) (_ bool, err error) {
```
golanglemonade, line 53:
```suggestion
func VerifyDerivedKey(dk, password string)  (bool, error) {
```

Same here

internal/passwd/errors.go
```diff
@@ -0,0 +1,56 @@
...
+	ErrCannotParseEncodedEK = errors.New("cannot parse encoded derived key, matched expression does not contain enough subgroups")
+
+	// ErrDKProtocol is returned when the derived key algorithm specified does not match the derived key algo part
+	ErrDKProtocol = fmt.Errorf("current code only works with the the dk protcol %q", dkAlg)
+
+	// ErrExpectedDKVersion is returned when the version part of the derived key does not match the specified version
+	ErrExpectedDKVersion = fmt.Errorf("expected %s version %d", dkAlg, argon2.Version)
+
+	// ErrDKParseSalt
+	ErrDKParseSalt = errors.New("could not parse salt: %s")
```
golanglemonade, line 34:
```suggestion
```

These aren't used I don't think

golanglemonade (approved):

golanglemonade (approved):

PR #131, opened 2023-11-26
add keygen package

matoszz:
adds basic keygen package for generating API keys

### DIFF
--- a/internal/keygen/keygen.go
+++ b/internal/keygen/keygen.go
@@ -1,0 +1,84 @@
+package keygen
+
+import (
+	"crypto/rand"
+	"encoding/binary"
+	"fmt"
+	"strings"
+)
+
+const (
+	alphabet = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"
+	alphanum = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
+	idxbits  = 6
+	idxmask  = 1<<idxbits - 1
+	idxmax   = 63 / idxbits
+)
+
+// Defaults for the length of key IDs and secrets in Datum
+const (
+	KeyIDLength  = 32
+	SecretLength = 64
+	ByteLength   = 8
+)
+
+// Alpha generates a random string of n characters that only includes upper and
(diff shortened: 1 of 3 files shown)


internal/keygen/keygen.go
```diff
@@ -0,0 +1,84 @@
...
+	idxbits  = 6
+	idxmask  = 1<<idxbits - 1
+	idxmax   = 63 / idxbits
+)
+
+// Defaults for the length of key IDs and secrets in Datum
+const (
+	KeyIDLength  = 32
+	SecretLength = 64
+	ByteLength   = 8
```
golanglemonade, line 22: Do these need to be exported? I assume they are just going to be referenced in this package?

internal/keygen/keygen_test.go
```diff
@@ -0,0 +1,119 @@
...
+		require.Regexp(t, regexp.MustCompile(`[a-zA-Z]+`), alpha)
+	}
+
+	vals := make(map[string]struct{})
+
+	for i := 0; i < 10000; i++ {
+		val := keygen.Alpha(16)
+		vals[val] = struct{}{}
+	}
+	require.Len(t, vals, 10000, "there is a very low chance that a duplicate value was generated")
```
golanglemonade, line 34: What is being tested here? That it can created 10000 unique values?

internal/keygen/keygen_test.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+func TestKeyGen(t *testing.T) {
+	// This is a long running test, skip if in short mode
+	if testing.Short() {
+		t.Skip("skipping long running test in short mode")
+	}
```
golanglemonade, line 64: This test is long running?

internal/keygen/keygen_test.go
matoszz: yea for collisions

internal/keygen/keygen_test.go
matoszz: the performance + enumeration tests are long running

internal/keygen/keygen.go
matoszz: I planned to use the keygen package for a few things and the lengths could change

golanglemonade (approved):

PR #132, opened 2023-11-26
add utils package

matoszz:
- adds a table writer / rows package for CLI use
- adds a new db driver interface for performing things like backups
- adds ulids utility for leveraging the oklog/ulid/v2 package which will be used in future PR's to create lexicographically sortable ID's associated with sessions, etc.


internal/utils/sqlite/sqlite.go
```diff
@@ -0,0 +1,116 @@
...
+
+import (
+	"database/sql"
+	"database/sql/driver"
+	"sync"
+
+	"github.com/mattn/go-sqlite3"
+)
+
+// Init creates the connections map and registers the driver with the SQL package.
```
golanglemonade, line 11:
```suggestion
// init creates the connections map and registers the driver with the SQL package.
```

internal/utils/sqlite/sqlite.go
```diff
@@ -0,0 +1,116 @@
...
+// Open implements the sql.Driver interface and returns a sqlite3 connection that can
+// be fetched by the user using GetLastConn. The connection ensures it's cleaned up
+// when it's closed. This method is not used by the user, but rather by sql.Open.
+func (d *Driver) Open(dsn string) (_ driver.Conn, err error) {
+	var inner driver.Conn
+
+	if inner, err = d.SQLiteDriver.Open(dsn); err != nil {
+		return nil, err
+	}
+
```
golanglemonade, line 46: Let's be consistent with our function signatures, theres no need to just name the error return

```suggestion
func (d *Driver) Open(dsn string) (driver.Conn, error) {
	inner, err := d.SQLiteDriver.Open(dsn); 
	if err != nil {
		return nil, err
	}
```

internal/utils/ulids/ulids.go
```diff
@@ -0,0 +1,121 @@
...
+	uid, err := ulid.New(ms, entropy)
+
+	if err != nil {
+		panic(err)
+	}
+
+	return uid
+}
+
+// FromTime ceates a ULID with the specified timestamp
```
golanglemonade, line 52:
```suggestion
// FromTime creates a ULID with the specified timestamp
```

internal/utils/ulids/ulids.go
```diff
@@ -0,0 +1,121 @@
...
+// Bytes parses a ULID and returns the []byte representation of the ULID
+func Bytes(uid any) (_ []byte, err error) {
+	var sid ulid.ULID
+
+	if sid, err = Parse(uid); err != nil {
+		return nil, err
+	}
+
+	return sid.Bytes(), nil
+}
```
golanglemonade, line 111:
```suggestion
func Bytes(uid any) ([]byte, error) {
	sid, err := Parse(uid)
	if err != nil {
		return nil, err
	}

	return sid.Bytes(), nil
}
```

internal/utils/ulids/ulids.go
```diff
@@ -0,0 +1,121 @@
...
+
+// MustBytes parses the ULID into bytes but panics on errors
+func MustBytes(uid any) (id []byte) {
+	var err error
+	if id, err = Bytes(uid); err != nil {
+		panic(err)
+	}
+
+	return id
+}
```
golanglemonade, line 121:
```suggestion
func MustBytes(uid any) []byte {
	id, err := Bytes(uid)
	if err != nil {
		panic(err)
	}

	return id
}
```

golanglemonade (approved):

PR #133, opened 2023-11-26
Update auditmixin to set createdby and updatedby

### LINKED ISSUE #103: Set createdby and updatedby in AuditMixin
Currently objects created in the system do not have the createdby or updatedby set because it's missing from the AuditMixin

```
go run cmd/cli/main.go org create -n "Lemonade Golang" -d "a new org created with the CLI"
{
  "createOrganization": {
    "organization": {
      "createdAt": "2023-11-24T13:41:16.662749-07:00",
      "createdBy": "", <--- should be set
      "displayName": "unknown",
      "id": "2h8fQ-XNuqi0oQftLTKBZ",
      "name": "Lemonade Golang",
      "updatedAt": "2023-11-24T13:41:16.662729-07:00",
      "updatedBy": "" <--- should be set
    }
  }
}
```

tylerauerbeck:
This updates the auditmixin so that `createdBy` and `updatedBy` are set appropriately when auth is enabled. The actor ID is grabbed from the echo context and used as the value in the cases where it is set. If no value is able to be grabbed from the context, it is assumed that auth is not enabled and the value is set to an empty string. 

The assumption here is that we would fail before arriving at this code path in the cases where auth is configured, but no value would be grabbed from the echo context.

While fixing this, I also made the assumption that `createdBy` should be `Immutable`, as this doesn't seem like something you would want someone to be able to update after the fact. If this is something that would be better discussed outside of this PR or done as a follow up, let me know and I can revert.

Resolves #103


matoszz (approved):
```
➜  datum git:(update-auditmixin) ✗ go run cmd/cli/main.go org create -n don_gato_house_of_pain -d "i love you tyler"
{
  "createOrganization": {
    "organization": {
      "createdAt": "2023-11-26T01:32:21.223815-07:00",
      "createdBy": "mattisthebest",
      "description": "i love you tyler",
      "displayName": "unknown",
      "id": "yWpX6NIlxDx-azoUa_L_k",
      "name": "don_gato_house_of_pain",
      "updatedAt": "2023-11-26T01:32:21.223801-07:00",
      "updatedBy": "mattisthebest"
    }
  }
}
```

internal/ent/mixin/audit_mixin.go
```diff
@@ -56,11 +59,25 @@ func AuditHook(next ent.Mutator) ent.Mutator {
...
 
+		ec, err := echox.EchoContextFromContext(ctx)
+		if err != nil {
+			return nil, err
+		}
+
+		actor, err := echox.GetActorSubject(*ec)
+		if err != nil {
+			actor = ""
+		}
```
golanglemonade, line 70: I got tired to typing both of these every time, so there is also `GetUserIDFromContext` that you can use instead

PR #135, opened 2023-11-26
move hooks to its own package

golanglemonade:
This conforms hooks to the same pattern as interceptors and privacy rules, which are all self-contained packages within `ent` and referenced by the schemas.

### DIFF
--- a/internal/ent/hooks/organization.go
+++ b/internal/ent/hooks/organization.go
@@ -1,0 +1,90 @@
+package hooks
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"entgo.io/ent"
+
+	"github.com/datumforge/datum/internal/ent/generated"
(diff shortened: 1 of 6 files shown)


internal/ent/hooks/organization.go
```diff
@@ -0,0 +1,90 @@
...
+		if exists {
+			tuples, err := createTuple(ctx, &m.Authz, fga.OwnerRelation, object)
+			if err != nil {
+				return err
+			}
+
+			if _, err := m.Authz.CreateRelationshipTuple(ctx, tuples); err != nil {
+				m.Logger.Errorw("failed to create relationship tuple", "error", err)
+
+				// TODO: rollback mutation if tuple creation fails
```
matoszz, line 57: can you link to an issue instead?

PR #145, opened 2023-11-27
add cookie and session store

matoszz:
- implements a new `store` for cookies and sessions based on gorilla sessions
- creates basic CRUD for cookies
- creates basic CRUD for sessions
- should implement the ulid package for randomness related to `NewState()`


internal/store/doc.go
```diff
@@ -0,0 +1,2 @@
+// Package store is for storing cookies, sessions, and maybe other fun stuff
```
golanglemonade, line 1: `and maybe other fun stuff` :) I like it

internal/store/session.go
```diff
@@ -0,0 +1,51 @@
...
+		return "", err
+	}
+
+	return cookieValue, nil
+}
+
+// SetSession function sets a session value in a session store and also sets a corresponding cookie in the response
+func SetSession(r *http.Request, w http.ResponseWriter, value, cookieName, valueName string, sessionStore SessionStore) error {
+	// set the cookie
+	session, _ := sessionStore.Get(r, cookieName)
```
golanglemonade, line 35: Should check this error?

internal/store/session.go
```diff
@@ -0,0 +1,51 @@
...
+	session.Values[valueName] = value
+
+	if err := session.Save(r, w); err != nil {
+		return err
+	}
+
+	return nil
```
golanglemonade, line 42: nit: you can just return this

```suggestion
	return session.Save(r, w)
```

internal/store/session.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+// RemoveSession function removes a session from the session store based on the provided cookie name
+func RemoveSession(r *http.Request, w http.ResponseWriter, cookieName string, sessionStore SessionStore) error {
+	session, _ := sessionStore.Get(r, cookieName)
```
golanglemonade, line 47: Same here, error is being discarded ?

internal/store/state.go
```diff
@@ -0,0 +1,20 @@
...
+)
+
+// NewState generates cryptographically secure random state with base64 URL encoding
+func NewState() (string, error) {
+	cryptoRandBytes := make([]byte, 16) //nolint:gomnd
+	_, err := rand.Read(cryptoRandBytes)
+
+	if err != nil {
+		return "", err
+	}
```
golanglemonade, line 15: nit: cuddle the err check with the definition 
```suggestion
	cryptoRandBytes := make([]byte, 16) //nolint:gomnd
	
	_, err := rand.Read(cryptoRandBytes)
	if err != nil {
		return "", err
	}
```

golanglemonade (approved):
just a couple small comments, otherwise LGTM

internal/store/session.go
```diff
@@ -0,0 +1,55 @@
...
+	// set the cookie
+	session, _ := sessionStore.Get(r, cookieName)
+	session.Values[valueName] = value
+
+	if err := session.Save(r, w); err != nil {
+		return err
+	}
+
+	return session.Save(r, w)
+}
```
golanglemonade, line 43: I think my two comments got confused, I meant the first session.Get isn't checking the error, like so: 

```suggestion
func SetSession(r *http.Request, w http.ResponseWriter, value, cookieName, valueName string, sessionStore SessionStore) error {
	// set the cookie
	session, err := sessionStore.Get(r, cookieName)
	if err != nil {
		return err
	}
	
	session.Values[valueName] = value

	return session.Save(r, w)
}
```

internal/store/session.go
```diff
@@ -0,0 +1,55 @@
...
+func RemoveSession(r *http.Request, w http.ResponseWriter, cookieName string, sessionStore SessionStore) error {
+	session, _ := sessionStore.Get(r, cookieName)
+	session.Options.MaxAge = -1
+
+	if err := session.Save(r, w); err != nil {
+		return err
+	}
+
+	return session.Save(r, w)
+}
```
golanglemonade, line 55: Same here: 

```suggestion
func RemoveSession(r *http.Request, w http.ResponseWriter, cookieName string, sessionStore SessionStore) error {
	session, err := sessionStore.Get(r, cookieName)
	if err != nil {
		return err
	}
	
	session.Options.MaxAge = -1

	return session.Save(r, w)
}
```

golanglemonade (approved):

PR #147, opened 2023-11-28
viperconfig and basic cleanup

matoszz:
- adds doc.go to several `/internal/` packages 
- adds a viperconfig util to allow for more "distributed" flag binding

Note: PR was modified from original scope that included adding oauth; removing this from this PR for now and creating issue for followup / better implementation


internal/auth/oauth.go
```diff
@@ -0,0 +1,179 @@
...
+		if provider == Google {
+			return google.Endpoint
+		}
+
+		return github.Endpoint
+	}
+
+	if provider != Google && provider != Github {
+		return nil, newReadError("provider", provider, err)
+	}
```
golanglemonade, line 58: This should be checked first, all your checks above are assuming you have a valid provider. I would also suggest, since we will probably end up with more providers eventually, you write a function that checks if the provider is in `knownProviders`, e.g. 

```
func validProvider(provider string) bool {
	for _, p := range knownProviders {
		if p == provider {
			return true
		}
	}

	return false
}
```

internal/auth/oauth.go
```diff
@@ -0,0 +1,179 @@
...
+		return []string{"user:email", "repo", "read:packages", "write:packages", "workflow", "read:org"}
+	}
+
+	endpoint := func(provider string) oauth2.Endpoint {
+		if provider == Google {
+			return google.Endpoint
+		}
+
+		return github.Endpoint
+	}
```
golanglemonade, line 54: This works for now, but as mentioned in my other comment, we might have more oauth providers at some point and then these inline functions get large. I would suggest making them named functions instead and just call them from here.

internal/auth/oauth.go
```diff
@@ -0,0 +1,179 @@
...
+)
+
+var knownProviders = []string{Google, Github}
+
+// NewOAuthConfig creates a new OAuth2 config for the given provider and whether the client is a CLI or web client
+func NewOAuthConfig(provider string, cli bool) (*oauth2.Config, error) {
+	var err error
+
+	redirectURL := func(provider string, cli bool) string {
+		base := viper.GetString(fmt.Sprintf("%s.redirect_uri", provider))
```
golanglemonade, line 32: Where is this viper flag coming from? What happens if its not set, is there a default?

internal/auth/oauth.go
```diff
@@ -0,0 +1,179 @@
...
+// ValidateProviderToken validates the given token for the given provider
+func ValidateProviderToken(_ context.Context, provider string, token string) error {
+	var err error
+
+	if provider == Github {
+		// Create an OAuth2 token source with the PAT
+		tokenSource := oauth2.StaticTokenSource(&oauth2.Token{AccessToken: token})
+
+		// Create an authenticated GitHub client
+		oauth2Client := oauth2.NewClient(context.Background(), tokenSource)
```
golanglemonade, line 142: Do you want to pass that context in and use it? 

```suggestion
func ValidateProviderToken(ctx context.Context, provider string, token string) error {
	var err error

	if provider == Github {
		// Create an OAuth2 token source with the PAT
		tokenSource := oauth2.StaticTokenSource(&oauth2.Token{AccessToken: token})

		// Create an authenticated GitHub client
		oauth2Client := oauth2.NewClient(ctx, tokenSource)
```

internal/auth/oauth.go
```diff
@@ -0,0 +1,179 @@
...
+
+const (
+	Google = "google"
+	Github = "github"
+)
+
+var knownProviders = []string{Google, Github}
+
+// NewOAuthConfig creates a new OAuth2 config for the given provider and whether the client is a CLI or web client
+func NewOAuthConfig(provider string, cli bool) (*oauth2.Config, error) {
```
golanglemonade, line 28: Can you add a test for this ?

golanglemonade (approved):

PR #156, opened 2023-11-30
add basic caching, using entcache, to the db layer

golanglemonade:
Looking into possible reasons why queries are happening multiple times on a request. This adds a basic driver-level caching (with a TTL of 1 minute) to our db connections. 

There are more advanced configs we can setup, like a context level cache instead, but this looks like a good place to start. 

references: 
https://entgo.io/blog/2021/10/14/introducing-entcache/
https://github.com/ariga/entcache/blob/master/README.md


internal/entdb/client.go
```diff
@@ -94,6 +107,12 @@ func (c *EntClientConfig) NewMultiDriverDBClient(ctx context.Context, opts []ent
 		return nil, err
 	}
 
+	// Decorates the sql.Driver with entcache.Driver on the primaryDB
+	drv := entcache.NewDriver(
+		primaryDB,
```
matoszz, line 112: any reason not to include both?

```suggestion
		primaryDB,
		secondaryDB,
```

matoszz (approved):

internal/entdb/client.go
golanglemonade: Only the primary is used for reads, so didn’t think was needed on secondary. Although it won’t hurt it either.

golanglemonade: Fixed

matoszz (approved):

Issue #152, opened 2023-11-28
Add soft delete mixin


matoszz:
Add a soft delete mixin and add mixin to existing schema; mixin should have / set records with deleted information (deleted by, deleted date) and should prevent hard deletion from the system unless done through administrative actions.

By default the system should behave in such a way that records flagged for deletion are not hard-removed from the system but rather flagged (or moved into a different corresponding "deleted" table (e.g. deleted_orgs)). We will eventually also have to be able to apply "holds" to record deletion (in the event we are legally obliged to retain the record due to a subpoena or similar) which prevent "hard deletion" unless the hold is removed. 

Consideration / research should be done around right to remove related to GDPR (how we handle the requests, how they are processed, how we provide evidence which satisfies we are no longer retaining information, etc.)

tylerauerbeck:
Soft deletes are in place after #157 

The next two steps would be some "hold" pattern that would prevent hard deletes. We could continue following the mixing pattern, which could have add a `record_hold` field that we could add to the various schemas.

The last bit is figuring out _how_ to handle hard deletes. Do we want this as a separate resolver that would require elevated/admin/some other set of permissions on the user to trigger a hard delete? Or just the same delete resolver and an additional flag (but still elevated/admin/some other set of permissions)

Or something completely different. More just thinking out loud before working on the rest of this.

 cc/ @matoszz @golanglemonade

PR #157, opened 2023-11-30
Add initial soft delete mixin

tylerauerbeck:
This adds the initial soft delete mixin. This marks a change in behavior as any schema that uses this mixin will not trigger an actual delete , but will instead be intercepted and treated as an update. This in turn will set the `deleted_at` and `deleted_by` fields as appropriate.

This only partially addresses #152 as it will still need to be decided how to:
- trigger actual deletes
- mark records as held which will prevent actual deletes from taking place on those records


matoszz:
@tylerauerbeck it's failing almost immediately for me when I try and run it as well; i tried swapping in new token secrets, but no dice - will look into it further with @golanglemonade. In your branch you can remove / change that actions workflow to not run for now as it's not mandatory CI check to complete; the buildkite datum one is more important and the atlas bits can also be accomplished locally

internal/api/organization_test.go
```diff
@@ -453,6 +454,25 @@ func TestMutation_DeleteOrganization(t *testing.T) {
 
 			// make sure the deletedID matches the ID we wanted to delete
 			assert.Equal(t, tc.orgID, resp.DeleteOrganization.DeletedID)
+
+			// make sure the org isn't returned
+			mockCheckAny(mockCtrl, mc, reqCtx, true)
```
golanglemonade, line 459: it's worthing noting (although I still think this is a good check) that permissions are removed when the organization is deleted, so the `Check` to FGA would/should return `false` and access would be denied to the resource.

golanglemonade (approved):

PR #197, opened 2023-12-07
adds db healthcheck, moves to server opts

golanglemonade:
When all the things are enabled we now get: 

```
curl localhost:17608/ready  
{
  "fga": "OK",
  "sqlite_db_primary": "OK",
  "sqlite_db_secondary": "OK"
}
```

`fga` and `sqlite_db_secondary` are dependent on `auth=enabled` and `db-multi-write=enabled` respectively

### DIFF
--- a/internal/entdb/client.go
+++ b/internal/entdb/client.go
@@ -14,6 +14,7 @@ import (
 	ent "github.com/datumforge/datum/internal/ent/generated"
 	"github.com/datumforge/datum/internal/ent/interceptors"
 	"github.com/datumforge/datum/internal/httpserve/config"
+	"github.com/pkg/errors"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


internal/entdb/client.go
```diff
@@ -139,3 +154,14 @@ func (c *EntClientConfig) createSchema(ctx context.Context, db *entsql.Driver) e
 
 	return nil
 }
+
+// Healthcheck pings the DB to check if the connection is working
+func Healthcheck(client *entsql.Driver) func(ctx context.Context) error {
```
matoszz, line 159: remember we're not going to want our services to crashloopbackoff if the connection to the db is down / health checks are failing

matoszz (approved):

internal/entdb/client.go
golanglemonade: ack, this just defines the checks. We can just use this for startup to make sure we connect successfully, and not use it or a readiness probe.

PR #198, opened 2023-12-07
Tokenmanager

matoszz:
- created new `internal/tokens` package for the creation, signing, rotation, and verification of access tokens, refresh tokens, and session tokens
- updated the sessions, user, access_tokens, and refresh_tokens schemas
- created jwks cache + cache refresher
- created a /config in the root directory containing portions of the tokens + auth configs
- created (but not finished) the bulk of the auth middleware
- refactored / rewrote the `/store/sessions` setup and will likely do the same for the others; created new db client and begun to stub out the structure for CRUD operations around tokens


config/doc.go
```diff
@@ -0,0 +1,2 @@
+// Package config holds the configuration options for the datum server
```
golanglemonade, line 1: Why is this in the root? We also now have two packages in the same repo called `config`

internal/ent/schema/ohauthtootoken.go
```diff
@@ -0,0 +1,66 @@
...
+
+	"entgo.io/contrib/entgql"
+	"entgo.io/ent"
+	"entgo.io/ent/schema"
+	"entgo.io/ent/schema/field"
+
+	"github.com/datumforge/datum/internal/ent/mixin"
+)
+
+// OhAuthTooToken holds the schema definition for the OhAuthTooToken entity
```
golanglemonade, line 14: `OhAuthTooToken` 😆

internal/ent/schema/accesstoken.go
```diff
@@ -0,0 +1,76 @@
...
+		field.Time("expires_at").
+			Default(func() time.Time { return time.Now().Add(time.Hour * 24 * 7) }), // nolint: gomnd
+		field.Time("issued_at").
+			Default(time.Now()),
+		field.Time("last_used_at").
+			UpdateDefault(time.Now).
+			Optional().
+			Nillable(),
+		field.String("organization_id").
+			Comment("organization ID of the organization the user is accessing"),
```
golanglemonade, line 36: Will this mean that I'm issued a new access token per org? This seems like it could be clunky when the user probably has a minimum of two orgs (personal org + another(1->many))

internal/ent/schema/refreshtoken.go
```diff
@@ -20,38 +19,23 @@ type RefreshToken struct {
...
+			Comment("organization ID of the organization the user is accessing"),
+		field.String("user_id").
+			Comment("the user the session is associated with"),
 	}
 }
 
 // Edges of the RefreshToken
 func (RefreshToken) Edges() []ent.Edge {
-	return []ent.Edge{
-		edge.From("user", User.Type).Ref("refreshtoken").Unique(),
```
golanglemonade, line 53: Why do you not need an edge to `user` here like you do in access token?

internal/httpserve/middleware/auth/authoptions.go
```diff
@@ -0,0 +1,393 @@
...
+// AuthOptions is constructed from variadic AuthOption arguments with reasonable defaults.
+type AuthOptions struct {
+	KeysURL            string           // The URL endpoint to the JWKS public keys on the datum server
+	Audience           string           // The audience to verify on tokens
+	Issuer             string           // The issuer to verify on tokens
+	MinRefreshInterval time.Duration    // Minimum amount of time the JWKS public keys are cached
+	CookieDomain       string           // The domain to use for auth cookies
+	Context            context.Context  // The context object to control the lifecycle of the background fetch routine
+	validator          tokens.Validator // The validator constructed by the auth options (can be directly supplied by the user).
+	reauth             Reauthenticator  // The refresher constructed by the auth options (can be directly supplied by the user).
```
golanglemonade, line 51: Put these comments above in the `KeysURL endpoint to...` format so they are picked up correctly for go docs

internal/httpserve/middleware/auth/authoptions.go
```diff
@@ -0,0 +1,393 @@
...
+func Reauthenticate(conf AuthOptions, validator tokens.Validator) func(c *echo.Context) (string, error) {
+	// If no reauthenticator is available on the configuration, always return an error.
+	if conf.reauth == nil {
+		return func(c *echo.Context) (string, error) {
+			return "", ErrRefreshDisabled
+		}
+	}
+
+	// If the reauthenticator is available, return a function that utilizes it.
+	return func(c *echo.Context) (_ string, err error) {
```
golanglemonade, line 253:
```suggestion
	return func(c *echo.Context) (string,  error) {
```

This is odd to have one named var on the return and one not

internal/httpserve/middleware/auth/authoptions.go
```diff
@@ -0,0 +1,393 @@
...
+	rftc := c.Cookie(RefreshTokenCookie)
+	if tks, err = c.Cookie(RefreshTokenCookie); err != nil {
+		return "", ErrNoRefreshToken
+	}
+	return tks, nil
+}
+
+// GetClaims fetches and parses datum claims from the echo context. Returns an
+// error if no claims exist on the context; panics if the claims are not the correct
+// type -- however the panic should be recovered by middleware.
```
golanglemonade, line 326: Why rely on the recover middleware instead of properly checking and returning an error instead of panic?

internal/tokens/claims.go
```diff
@@ -0,0 +1,53 @@
...
+type Claims struct {
+	jwt.RegisteredClaims
+	// UserID is the internal generated ID for the user
+	UserID string `json:"user_id,omitempty"`
+	// the email associated with the user
+	Email string `json:"email,omitempty"`
+	// the organization ID the JWT token is valid for
+	OrgID string `json:"org,omitempty"`
+	// the ID of the parent organization, if a child
+	ParentOrgID string `json:"parentorg,omitempty"`
```
golanglemonade, line 20: Does this imply the user will have access to the parent org, or just that one exists for their org?

internal/tokens/doc.go
```diff
@@ -0,0 +1,2 @@
+// Package tokens does token stuff
```
golanglemonade, line 1: yeah, you should update that one :)

internal/tokens/errors.go
```diff
@@ -0,0 +1,176 @@
...
+)
+
+// Error constants
+var (
+	ErrInvalidKey      = errors.New("key is invalid")
+	ErrInvalidKeyType  = errors.New("key is of invalid type")
+	ErrHashUnavailable = errors.New("the requested hash function is unavailable")
+
+	ErrCacheMiss    = errors.New("requested key is not in the cache")
+	ErrCacheExpired = errors.New("requested key is expired")
```
golanglemonade, line 15: comment these too?

internal/tokens/errors.go
```diff
@@ -0,0 +1,176 @@
...
+
+	// ErrTokenManagerFailedInit returns when the token manager was not correctly provided signing keys
+	ErrTokenManagerFailedInit = errors.New("token manager not initialized with signing keys")
+
+	// ErrFailedRetreiveClaimsFromToken returns when claims can not be retreived from an access token
+	ErrFailedRetreiveClaimsFromToken = errors.New("could not retrieve claims from access token")
```
golanglemonade, line 21:
```suggestion
	// ErrFailedRetrieveClaimsFromToken returns when claims can not be retrieved from an access token
	ErrFailedRetrieveClaimsFromToken = errors.New("could not retrieve claims from access token")
```

internal/tokens/tokenmanager.go
```diff
@@ -0,0 +1,390 @@
...
+
+	// Add the kid to the header
+	token.Header["kid"] = tm.currentKeyID.String()
+
+	// Return the signed string
+	return token.SignedString(tm.currentKey)
+}
+
+// CreateTokenPair returns signed access and refresh tokens for the specified claims in one step since usually you want both access and refresh tokens at the same time
+func (tm *TokenManager) CreateTokenPair(claims *Claims) (accessToken, refreshToken string, err error) {
```
golanglemonade, line 163:
```suggestion
func (tm *TokenManager) CreateTokenPair(claims *Claims) (string, string, error) {
```

Either don't name them like this^ or you should just do `return` instead of
```
return accessToken, refreshToken, nil
```

internal/tokens/tokenmanager.go
```diff
@@ -0,0 +1,390 @@
...
+
+	if accessToken, err = tm.Sign(atk); err != nil {
+		return "", "", fmt.Errorf("could not sign access token: %w", err)
+	}
+
+	if refreshToken, err = tm.Sign(rtk); err != nil {
+		return "", "", fmt.Errorf("could not sign refresh token: %w", err)
+	}
+
+	return accessToken, refreshToken, nil
```
golanglemonade, line 182: since you defined your return vars, you can just do this if you want

```suggestion
	return 
```

internal/tokens/tokenmanager.go
```diff
@@ -0,0 +1,390 @@
...
+	return tm.refreshAudience
+}
+
+// CurrentKey returns the ulid of the current key being used to sign tokens - this is just the identifier of the key, not the key itself
+func (tm *TokenManager) CurrentKey() ulid.ULID {
+	return tm.currentKeyID
+}
+
+// keyFunc selects the RSA public key from the list of tokenmanager internal keys based on the kid in the token header - if the kid does not exist an error is returned the token is not validated
+func (tm *TokenManager) keyFunc(token *jwt.Token) (key interface{}, err error) {
```
golanglemonade, line 309: This is basically the same function as [here](https://github.com/datumforge/datum/pull/198/files#diff-a0bacda97d952202cd5203bd965f2fb187befa6c7ba59889484556aa7272c74aR37). Why do we need both?

internal/tokens/tokens_test.go
```diff
@@ -0,0 +1,381 @@
+package tokens_test
```
golanglemonade, line 1: This is a good set of tests for everything together, but each function should really have its own unit test as well

internal/ent/schema/accesstoken.go
matoszz: it means we'll issue an access token for the org you're attempting to access

internal/httpserve/middleware/auth/authoptions.go
matoszz: same as above :)

internal/tokens/claims.go
matoszz: it doesn't imply anything; storing it in the context just is a helper for us

internal/tokens/tokenmanager.go
matoszz: because they're for different things. One is for public jwks URL and the other is for RSA keys we've loaded

internal/ent/schema/refreshtoken.go
matoszz: Refresh tokens are related only to access tokens, not to users

golanglemonade (approved):

PR #202, opened 2023-12-08
user hook, avatar, pass

matoszz:
- creates a user hook to argon2id encrypt the password and persist it in the db
- user hook also creates a hash of the user's email and generates a gravatar image based on that
- adds gravatar package to /utils/
- updates CLI to take password as an input parameter
- updates user graphql queries to blow out additional fields


internal/utils/gravatar/gravatar.go
```diff
@@ -0,0 +1,81 @@
...
+	"path/filepath"
+	"strconv"
+	"strings"
+)
+
+// Default values for gravatar options generated by this package.
+const (
+	DefaultSize   = 80
+	DefaultImage  = "robohash"
+	DefaultRating = "pg"
```
golanglemonade, line 16: I dont think these need to be exported

internal/ent/schema/user.go
```diff
@@ -159,3 +163,27 @@ func (User) Policy() ent.Policy {
...
+					if err != nil {
+						return nil, err
+					}
+
+					mutation.SetPassword(hash)
+				}
+
+				if email, ok := mutation.Email(); ok {
+					url := gravatar.New(email, nil)
+					mutation.SetAvatarRemoteURL(url)
```
golanglemonade, line 182: This should probably first check that one isn't already set by the user?

internal/utils/gravatar/gravatar_test.go
```diff
@@ -0,0 +1,22 @@
...
+
+func TestGravatar(t *testing.T) {
+	email := "sfunk@datum.net "
+	url := gravatar.New(email, nil)
+	require.Equal(t, "https://www.gravatar.com/avatar/241e2e069287a36e9c902e55a3b6af8a?d=robohash&r=pg&s=80", url)
+}
+
+func TestHash(t *testing.T) {
+	// Test case from: https://en.gravatar.com/site/implement/hash/
+	input := "sfunk@datum.net "
```
golanglemonade, line 19: idk if this will break your test, but there is a trailing space
```suggestion
	input := "sfunk@datum.net"
```

internal/utils/gravatar/gravatar_test.go
```diff
@@ -0,0 +1,22 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/datumforge/datum/internal/utils/gravatar"
+)
+
+func TestGravatar(t *testing.T) {
+	email := "sfunk@datum.net "
```
golanglemonade, line 12: Here too 
```suggestion
	email := "sfunk@datum.net"
```

internal/utils/gravatar/gravatar.go
matoszz: They do if you want to override them in the options

internal/ent/schema/user.go
matoszz: The email, or the avatar? The intent was to execute the hook on user creation and if an email wasn't already present; circumstantially there's no way a user could have set an avatar if the hook only executes on creation or if a user is updated without an email?

golanglemonade (dismissed):

PR #210, opened 2023-12-11
url tokens

matoszz:
- adds verification, register token types that are URL safe to send to users via email


internal/tokens/urltokens.go
```diff
@@ -0,0 +1,220 @@
...
+// the string can be recomputed when verifying a user provided token.
+func (t *VerificationToken) Sign() (_ string, secret []byte, err error) {
+	var data []byte
+
+	if data, err = msgpack.Marshal(t); err != nil {
+		return "", nil, err
+	}
+
+	return t.signData(data)
+}
```
golanglemonade, line 59:
```suggestion
func (t *VerificationToken) Sign() (string, []byte, error) {
	var data []byte

	if data, err := msgpack.Marshal(t); err != nil {
		return "", nil, err
	}

	return t.signData(data)
}
```

you aren't using the name vars, so doesn't make sense to declare them

internal/tokens/urltokens.go
```diff
@@ -0,0 +1,220 @@
...
+// the string can be recomputed when verifying a user provided token
+func (t *ResetToken) Sign() (_ string, secret []byte, err error) {
+	var data []byte
+
+	if data, err = msgpack.Marshal(t); err != nil {
+		return "", nil, err
+	}
+
+	return t.signData(data)
+}
```
golanglemonade, line 122:
```suggestion
func (t *ResetToken) Sign() (string, []byte, error) {
	var data []byte

	if data, err := msgpack.Marshal(t); err != nil {
		return "", nil, err
	}

	return t.signData(data)
}
```

Same here

golanglemonade (approved):

PR #220, opened 2023-12-12
fix the db client nil pointer

golanglemonade:
Tested and confirmed we can pull data from the database with the ent client in the handlers with this fix


internal/httpserve/serveropts/option.go
```diff
@@ -231,9 +230,6 @@ func WithAuth(settings map[string]any) ServerOption {
 // WithReadyChecks adds readiness checks to the server
 func WithReadyChecks(c *entdb.EntClientConfig, f *fga.Client) ServerOption {
 	return newApplyFunc(func(s *ServerOptions) {
-		// Initialize checks
-		s.Config.Server.Handler = handlers.Handler{}
```
golanglemonade, line 235: Previously, this was cleared the entdbClient that was set in serve.

matoszz (approved):

PR #225, opened 2023-12-12
Adds login handler

golanglemonade:
Login handler takes a username (email) and password and confirms:
- the user is in the database
- the password matches
- the email is verified

Upon success, it adds the access and refresh token to the header

To test it out:

Run the server:
```
 task run-dev-auth
```

Create a user, and set the email-verified to true in the database

hit the login endpoint with username + pass
```
curl  -v localhost:17608/login -d '{"username":"sfunk@datum.net", "password":"this#is-notreal$123%"}'
```

The cookies will be set in the response.

Follow-up PR will add this to the cli as well.

### DIFF
--- a/internal/httpserve/handlers/handlers.go
+++ b/internal/httpserve/handlers/handlers.go
@@ -8,13 +8,18 @@ import (
 	"github.com/datumforge/datum/internal/tokens"
 )
 
-// Handler contains configuration options for handlers including ReadyChecks and JWTKeys
+// Handler contains configuration options for handlers
 type Handler struct {
 	// DBClient to interact with the generated ent schema
-	DBClient     *ent.Client
-	TM           *tokens.TokenManager
+	DBClient *ent.Client
+	// TM contains the token manager in order to validate auth requests
+	TM *tokens.TokenManager
+	// CookieDomain is the domain set in cookie for authenticated requests, defaults to datum.net
 	CookieDomain string
-	Logger       *zap.SugaredLogger
-	ReadyChecks  Checks
-	JWTKeys      jwk.Set
+	// Logger provides the zap logger to do logging things from the handlers
+	Logger *zap.SugaredLogger
+	// ReadyChecks is a set of checkFuncs to determine if the application is "ready" upon startup
+	ReadyChecks Checks
+	// JWTKeys contains the set of valid JWT authentication key
+	JWTKeys jwk.Set
 }
--- a/internal/httpserve/handlers/handlers.go
+++ b/internal/httpserve/handlers/handlers.go
@@ -2,14 +2,19 @@ package handlers
 
 import (
 	"github.com/lestrrat-go/jwx/v2/jwk"
+	"go.uber.org/zap"
 
 	ent "github.com/datumforge/datum/internal/ent/generated"
+	"github.com/datumforge/datum/internal/tokens"
 )
 
 // Handler contains configuration options for handlers including ReadyChecks and JWTKeys
 type Handler struct {
 	// DBClient to interact with the generated ent schema
-	DBClient    *ent.Client
-	ReadyChecks Checks
-	JWTKeys     jwk.Set
+	DBClient     *ent.Client
+	TM           *tokens.TokenManager
+	CookieDomain string
+	Logger       *zap.SugaredLogger
+	ReadyChecks  Checks
+	JWTKeys      jwk.Set
 }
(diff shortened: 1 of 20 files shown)


internal/httpserve/handlers/handlers.go
```diff
@@ -2,14 +2,24 @@ package handlers
...
 )
 
-// Handler contains configuration options for handlers including ReadyChecks and JWTKeys
+// Handler contains configuration options for handlers
 type Handler struct {
 	// DBClient to interact with the generated ent schema
-	DBClient    *ent.Client
+	DBClient *ent.Client
+	// TM contains the token manager in order to validate auth requests
+	TM *tokens.TokenManager
```
matoszz, line 16: Should this still be called just "handler" or make it more specific?

internal/httpserve/handlers/login.go
```diff
@@ -0,0 +1,96 @@
...
+		RegisteredClaims: jwt.RegisteredClaims{
+			Subject: u.userID,
+		},
+		UserID: u.userID,
+		Email:  u.Username,
+	}
+}
+
+// verifyCredentials verifies the username and password are valid
+func (h *Handler) verifyCredentials(ctx echo.Context) (*User, error) {
```
matoszz, line 57: I have many of these structs defined in my branch but denoted as "Passkey Credentials"; it might be worthwhile to make some of these more specific like "verifyUserPassword" so that when we blow out other authentication functions it's not confusing which purpose they're serving

internal/httpserve/handlers/login.go
```diff
@@ -0,0 +1,96 @@
...
+	// set cookies on request with the access and refresh token
+	// when cookie domain is localhost, this is dropped but expected
+	if err := auth.SetAuthCookies(ctx, access, refresh, h.CookieDomain); err != nil {
+		return auth.ErrorResponse(err)
+	}
+
+	return ctx.JSON(http.StatusOK, "success")
+}
+
+func createClaims(u *User) *tokens.Claims {
```
matoszz, line 46: Do we not have createClaims already defined elsewhere?

internal/httpserve/handlers/login.go
```diff
@@ -0,0 +1,96 @@
...
+	return &tokens.Claims{
+		RegisteredClaims: jwt.RegisteredClaims{
+			Subject: u.userID,
+		},
+		UserID: u.userID,
+		Email:  u.Username,
```
matoszz, line 52: UserName / camel case ?

internal/tokens/flags.go
```diff
@@ -32,6 +32,11 @@ func RegisterAuthFlags(v *viper.Viper, flags *pflag.FlagSet) error {
 		return err
 	}
 
+	err = viperconfig.BindConfigFlag(v, flags, "jwt.cookie-domain", "jwt-cookie-domain", "datum.net", "cookie domain for datum JWT", flags.String)
+	if err != nil {
```
matoszz, line 36: Should this be configurable ?

internal/httpserve/handlers/handlers.go
golanglemonade: Its used for all handlers, e.g. any handler that will need the database will call `h.DBClient`, its not necessarily specific to this one

internal/httpserve/handlers/login.go
golanglemonade: not specific to taking the user from the `POST` form and creating the claims. Its just returning the `tokens.Claims` struct.

internal/httpserve/handlers/login.go
golanglemonade: from what I've seen `username` is considered "one word" as opposed to "user name" so theres no reason to capitalize the `N` here

internal/tokens/flags.go
golanglemonade: As an open source project, I would assume so? We would also want it configurable for testing vs. production

internal/httpserve/handlers/handlers.go
golanglemonade: I thought about adding a login config as well, within Handler, but thats only really the cookie domain and maybe token manager (assuming other handlers aren't also going to require that). There are already other things like ReadyChecks that not *all* handlers need, but are required for one or more of the rest endpoints

matoszz (approved):

PR #226, opened 2023-12-12
upgrade openfga v0.3.0

golanglemonade:


internal/fga/fga.go
```diff
@@ -200,7 +200,7 @@ func (c *Client) CreateStore(ctx context.Context, storeName string) (string, err
 
 	// Only create a new test store if one does not exist
 	if len(stores.GetStores()) > 0 {
```
matoszz, line 202: does this need an error check?

matoszz (approved):

internal/fga/fga.go
golanglemonade: No, it doesn't return an error, will return 0 if there are no existing stores

PR #229, opened 2023-12-13
set durations for tokens

golanglemonade:
Without setting these values in the config, the `iat` == `exp`. This sets the defaults and allows them to be overridden

```
Token claims
------------
{
  ...
  "exp": 1702489302,
  "iat": 1702485702,
...
  "nbf": 1702485702,
...
}
```


internal/tokens/flags.go
```diff
@@ -11,6 +11,9 @@ import (
 
 const (
 	defaultJWKSRemoteTimeout = 5 * time.Second
+	defaultAccessDuration    = 1 * time.Hour
+	defaultRefreshDuration   = 2 * time.Hour
+	defaultRefreshOverlap    = -15 * time.Minute
```
matoszz, line 16: what's the circumstance you would want a longer or shorter length of refresh issuance?

golanglemonade: if the access + refresh durations are shorter, you'd also want a shorter refresh overlap (similar for them being longer)

matoszz: should there be a check to ensure these configured times don't create conflicts?

matoszz (approved):

internal/tokens/flags.go
golanglemonade: yeah, I was thinking about that when it wasn't set at all it ended up with a bad config. Added: https://github.com/datumforge/datum/issues/232

Issue #155, opened 2023-11-28
Split datum cli cmd into multiple folders


golanglemonade:
Each object (e.g organization, user, etc) should have its own package + root cmd.

This will first require exporting some of the functions such as `viperBindFlag` to be used in multiple packages. 

Current Structure:
```
cmd/cli
├── cmd
│   ├── errors.go
│   ├── org.go
│   ├── org_create.go
│   ├── org_delete.go
│   ├── org_get.go
│   ├── org_update.go
│   ├── root.go
│   ├── user.go
│   ├── userCmd
│   ├── user_create.go
│   ├── user_get.go
│   └── user_update.go
└── main.go
```

Expected Result: 
```
cmd/cli
├── cmd
│   ├── errors.go
│   ├── root.go
│   ├── orgCmd
│   │   ├── org.go
│   │   ├── org_create.go
│   │   ├── org_delete.go
│   │   ├── org_get.go
│   │   └── org_update.go
│   ├── userCmd
│   │   ├── user.go
│   │   ├── user_create.go
│   │   ├── user_get.go
│   │   └── user_update.go
...
└── main.go
```

PR #231, opened 2023-12-13
refactor cli layout, add self flag on user command

golanglemonade:
- Moves all cmds into their own package/directory. This now requires the cmd package to be imported in `main.go` but you can see the commands all still show up: 

```
./datum-cli --help         
the datum cli

Usage:
  datum [command]

Available Commands:
  completion  Generate the autocompletion script for the specified shell
  group       The subcommands for working with datum groups
  help        Help about any command
  org         The subcommands for working with the datum organization
  pat         The subcommands for working with personal access tokens
  user        The subcommands for working with the datum user
```

- Adds a `--self` flag to the `user get` command: 
```
./datum-cli user get --self
{
  "user": {
    "avatarRemoteURL": "https://www.gravatar.com/avatar/723bac4e520e0b11e057060e88b68ae8?d=robohash\u0026r=pg\u0026s=80",
    "createdAt": "2023-12-12T11:02:51.609546-07:00",
    "createdBy": "unknown",
    "displayName": "sfunk",
...
```
- Simplifies `errors.go` for missing required fields

Resolves https://github.com/datumforge/datum/issues/155


cmd/cli/cmd/errors.go
```diff
@@ -1,47 +1,29 @@
...
-	// ErrTokenNameRequired is returned when no token name is provided when creating a new token
-	ErrTokenNameRequired = errors.New("token name is required")
-
-	// ErrTokenIDRequired is returned when no token ID is provided when deleting an existing token
-	ErrTokenIDRequired = errors.New("token ID is required")
-)
+// RequiredFieldMissingError is returned when a field is required but not provided
+type RequiredFieldMissingError struct {
+	Field      string
+	ObjectType string
```
matoszz, line 16: comment on the exported fields?

cmd/cli/cmd/group/group_update.go
```diff
@@ -53,13 +54,13 @@ func updateGroup(ctx context.Context) error {
...
 	// new client with params
-	c := datumclient.NewClient(h, host, opt, i)
+	c := datumclient.NewClient(h, datum.GraphAPIHost, opt, i)
 
 	var s []byte
 
 	oID := viper.GetString("group.update.id")
 	if oID == "" {
-		return ErrGroupIDRequired
+		return datum.NewRequiredFieldMissingError("group id")
```
matoszz, line 63: should you include the actual ID in the error response or does the "group ID" actually translate to the ID via the ObjectType?

cmd/cli/cmd/pat/pat.go
```diff
@@ -0,0 +1,18 @@
+// Package datumpat is our cobra/viper cli for personal access token endpoints
```
matoszz, line 1: let's stash this into a `tokens` directory so that as we add additional commands around issuance / testing in the CLI they're already grouped?

cmd/cli/cmd/root.go
```diff
@@ -21,37 +21,38 @@ const (
 
 var (
 	cfgFile string
-	host    string
 	logger  *zap.SugaredLogger
+	// GraphAPIHost contains the url for the graph api
```
matoszz, line 25: would this ever be different than the http host?

cmd/cli/cmd/user/user.go
```diff
@@ -1,6 +1,11 @@
-package datum
+// Package datumuser is our cobra/viper cli for user endpoints
```
matoszz, line 1: mostly a nit, but we use `doc.go` everywhere else

internal/httpserve/middleware/auth/user.go
```diff
@@ -9,8 +9,8 @@ import (
 	"github.com/datumforge/datum/internal/utils/ulids"
 )
 
-// GetActorSubject returns the user from the echo.Context
-func GetActorSubject(c echo.Context) (string, error) {
+// GetActorUserID returns the user from the echo.Context
+func GetActorUserID(c echo.Context) (string, error) {
```
matoszz, line 13: do we need this in both `ent/hooks/tuples/go` and `httpserve/middleware/auth/user.go`?

matoszz (dismissed):

cmd/cli/cmd/group/group_update.go
golanglemonade: well its missing, this is the same error message as before saying `group id is required` , meaning we don't have the ID

cmd/cli/cmd/root.go
golanglemonade: Should be the root + query, on my current branch I'm updating this to interact with rest endpoints so I'm going to resolve this for now and we can address in my next PR

internal/httpserve/middleware/auth/user.go
golanglemonade: tuples.go uses this function

cmd/cli/cmd/pat/pat.go
golanglemonade: will move in a follow-up PR

cmd/cli/cmd/user/user.go
golanglemonade: will update in a follow-up PR

cmd/cli/cmd/errors.go
golanglemonade: yeah, see will do in a new PR as well :)

Issue #168, opened 2023-12-01
Ensure cascade deletes work with soft-deletes


golanglemonade:
We recently enabled the soft-delete mixin which uses an ent hook that changes the operation from a delete to an update. 

```
    sd.SetOp(ent.OpUpdate)
```
As part of this change, we need to ensure cascade deletes still work, or update the code to allow these to work.

Example: 
https://github.com/datumforge/datum/blob/main/internal/ent/schema/organization.go#L87-L89

```
edge.To("groups", Group.Type).Annotations(entsql.Annotation{OnDelete: entsql.Cascade}),
edge.To("integrations", Integration.Type).Annotations(entsql.Annotation{OnDelete: entsql.Cascade}),
edge.To("setting", OrganizationSetting.Type).Unique().Annotations(entsql.Annotation{OnDelete: entsql.Cascade}),
```

matoszz:
[PR251](https://github.com/datumforge/datum/pull/251) addresses the bulk of the functionality required, but currently the implemented code path performs an exec and returns an err which will create orphaned records. The remaining functionality which would make this feature complete would be to evaluate the transaction and perform a rollback when encountering an error instead of an exec + error.

PR #251, opened 2023-12-16
Add initial edgecleanup helper functions

tylerauerbeck:
After adding soft deletes, it was noticed that cascading deletes stopped working as expected per: https://github.com/datumforge/datum/issues/168

To handle for this, this PR adds a template and annotation that we can use to properly trigger the cascading delete for related records in a schemas edge(s).

The annotation can be added to an edge like this: 
```go
edge.To("groups", Group.Type).Annotations(entsql.Annotation{OnDelete: entsql.Cascade}, entx.CascadeAnnotationField("Owner"))
```

Once that is added, it will be picked up by the new template that has been added. For each edge with the annotation a delete call will be added to a function named according to the schema it was applied to (i.e. if organization has an edge named group with the annotation, it will generate a func called `organizationEdgeCleanup` with an action inside it that calls 
```go
_ = r.Group.Delete().Where(group.HasOwnerWith(organization.ID(id))).ExecX(ctx)
```

This will then explicitly call delete on records with the matching ID, triggering the cascading delete (with the cascading delete triggering the soft delete if the mixin is configured). This was confirmed manually and tests have been added to confirm this as well.

Ex:
```
go run cmd/cli/main.go org create -n "tyler" -d "skynet devel" --no-auth
go run cmd/cli/main.go group create  -n notangryrobots -d "just some robots, nothing sketchy" -o <orgID> -s notangryrobo --no-auth
go run cmd/cli/main.go org delete -i <orgID> --no-auth
```

Once you've run these commands, you can confirm that the group associated with the deleted org was soft deleted with
```
task db:console
sqlite> select id, deleted_at from groups;
┌────────────────────────────┬──────────────────────────────────┐
│             id             │            deleted_at            │
├────────────────────────────┼──────────────────────────────────┤
│ 01HHRGGRZZ5PT5JK8FK06AQT3Y │ 2023-12-16 00:11:33.942499-05:00 │
└────────────────────────────┴──────────────────────────────────┘
```


internal/ent/templates/edge_cleanup.tmpl
```diff
@@ -0,0 +1,26 @@
...
+    {{/* For each schema */}}
+	{{- range $node := $.Nodes }}
+            {{/* create an EdgeCleanup function accepting a client and an ID */}}
+			func {{ $node.Name }}EdgeCleanup(ctx context.Context, r *Client, id string) error {
+               {{/* For each edge */}}
+               {{- range $edge := $node.Edges }}
+                    {{/* if the edge has our custom annotation applied */}}
+                    {{- if $annotation := $edge.Annotations.DATUM_CASCADE }}
+                        {{/* use the client to delete records where the edge schema has a field (provided by the annotation) containing the ID provided by the func */}}
+                        _ = r.{{ $edge.Type.Name }}.Delete().Where({{ $edge.Type.Name | lower }}.Has{{ $annotation.Field }}With({{ $node.Name | lower }}.ID(id))).ExecX(ctx)
```
golanglemonade, line 19: This will panic if its not successful and doesn't return a value, but the function itself is allowing an error return.  Should this be `Exec` instead of `ExecX` and return the error like so? 

```suggestion
                        if err = r.{{ $edge.Type.Name }}.Delete().Where({{ $edge.Type.Name | lower }}.Has{{ $annotation.Field }}With({{ $node.Name | lower }}.ID(id))).Exec(ctx); err != nil {
                            return err
                        }
```

matoszz:
@tylerauerbeck this is awesome, thank you so much for finishing out the soft delete functionality! Have you tested the behavior of this with the existing built-in soft-delete setup? Meaning, are they mutually exclusive? Should we build in some checks / lints at a later date to ensure only 1 is used? Just curious about any conflicts with the SQL-based cascade functionality that ent employs today.

tylerauerbeck:
@matoszz So this actually works in conjunction with the soft delete hook I added, not replacing it. On the other hand, it does replace the need for the entsql cascade annotation, as we're now explicitly triggering the deletes and triggering the hooks on the edges that have our new annotation ( so we're doing cascades, but with 💪 ).

internal/ent/templates/edge_cleanup.tmpl
tylerauerbeck: Yeah, I was waiting to get transactions plumbed in through the rest of the calls and didn't want to overload this PR. Maybe we can at least return the `err` for now with a `TODO` of rolling back if an err is hit? The result here is if we error on a delete, we would end up with orphaned records that would need to be cleaned up.

matoszz: Noted the follow up on this in https://github.com/datumforge/datum/issues/168

golanglemonade (approved):

PR #265, opened 2023-12-18
add org setting get

golanglemonade:
Instead of getting distracted, just pushing this, which implements the `OrganizationSetting` `GET` resolvers. Will follow-up l later with the mutations


matoszz (dismissed):

query/organizationsetting.graphql
```diff
@@ -0,0 +1,19 @@
...
+    ssoEntrypoint
+    billingAddress
+    billingContact
+    billingEmail
+    billingPhone
+    createdAt
+    createdBy
+    domains
+    updatedAt
+    updatedBy
```
matoszz, line 17: I've been finding it useful to pass back the associated organization ID in the query for these types of objects

```suggestion
    updatedBy
    organization {
      id
    }
```

matoszz (dismissed):

matoszz (approved):

PR #266, opened 2023-12-18
Groupauthz

matoszz:


internal/ent/hooks/group.go
```diff
@@ -41,6 +46,98 @@ func HookGroup() ent.Hook {
...
+			tupleGroup = append(tupleGroup, orgTuples...)
+
+			if _, err := m.Authz.CreateRelationshipTuple(ctx, tupleGroup); err != nil {
+				m.Logger.Errorw("failed to create relationship tuple", "error", err)
+
+				return ErrInternalServerError
+			}
+		}
+
+		m.Logger.Infow("created admin relationship tuples", "relation", fga.AdminRelation, "object", object)
```
golanglemonade, line 110: I dont think we need this, its outside the if of any creation
```suggestion
```

internal/ent/privacy/rule/group.go
```diff
@@ -0,0 +1,137 @@
...
+// CanCreateGroupsInOrg is a rule that returns allow decision if user has edit access in the organization
+func CanCreateGroupsInOrg() privacy.GroupMutationRuleFunc {
+	return privacy.GroupMutationRuleFunc(func(ctx context.Context, m *generated.GroupMutation) error {
+		gCtx := graphql.GetFieldContext(ctx)
+
+		oID, ok := gCtx.Args["owner"].(string)
+		if !ok {
+			return privacy.Allowf("nil request, bypassing auth check")
+		}
+		m.Logger.Debugw("checking mutation access")
```
golanglemonade, line 104: This isn't an argument on a group mutation, so it will always drop into `privacy.Allow()`, the group will get created, but then cant be retrieved

PR #270, opened 2023-12-19
Generate sanity check

matoszz:
This PR started generally as a desire to perform a CI-based sanity check on the `generate` portions of the codebase so that if one were to take a suggestion / commit as a part of a PR review, if the committed changes require re-generation of the schema and/or client scripts, this check should fail the buildkite "pre-check" step.

When testing this functionality, although it worked intermittently, there was a number of builds which failed due to the [generated schema](https://github.com/datumforge/datum/blob/main/internal/ent/generated/internal/schema.go) being regenerated. Upon inspection, it seemed as though the `defaultFunc` time generating functionality configured in several schemas against the field `expires_at` causing an updated dateTime entry in the field (and thus creating a change and causing the generation check to fail). Leveraging `onCreate` hooks seemed like a viable alternative over the `DefaultFunc` in the field so that it would only execute when the resource was created, hence creating hooks for accessTokens, sessions, etc.


internal/ent/hooks/session.go
```diff
@@ -0,0 +1,27 @@
...
+			if mutation.Op().Is(ent.OpCreate) {
+				expires, _ := mutation.ExpiresAt()
+				if expires.IsZero() {
+					mutation.SetExpiresAt(time.Now().Add(time.Hour * 24 * 7)) // nolint: gomnd
+				}
+			}
+
+			return next.Mutate(ctx, mutation)
+		})
+	}, ent.OpCreate|ent.OpUpdateOne)
```
golanglemonade, line 26: This should be able to just be `ent.OpCreate`  on line 26 and remove the `if` check

golanglemonade, line 26:
```suggestion
		return hook.SessionFunc(func(ctx context.Context, mutation *generated.SessionMutation) (generated.Value, error) {
		        expires, _ := mutation.ExpiresAt()
		        if expires.IsZero() {
			        mutation.SetExpiresAt(time.Now().Add(time.Hour * 24 * 7)) // nolint: gomnd
		        }

			return next.Mutate(ctx, mutation)
		})
	}, ent.OpCreate)
```

internal/ent/hooks/personalaccesstoken.go
```diff
@@ -0,0 +1,23 @@
...
+	"entgo.io/ent"
+
+	"github.com/datumforge/datum/internal/ent/generated"
+	"github.com/datumforge/datum/internal/ent/generated/hook"
+)
+
+// HookPersonalAccessToken runs on accesstoken mutations and sets expires
+func HookPersonalAccessToken() ent.Hook {
+	return func(next ent.Mutator) ent.Mutator {
+		return hook.PersonalAccessTokenFunc(func(ctx context.Context, m *generated.PersonalAccessTokenMutation) (ent.Value, error) {
```
golanglemonade, line 16: If we don't expect to need other things on this hook, We should set this up the same way as the session hook and remove the `if` and just have the entire hook only trigger `ent.OnCreate`

internal/ent/schema/entitlement.go
```diff
@@ -18,15 +18,27 @@ type Entitlement struct {
...
+			Default("free"),
+		field.String("external_customer_id").
+			Comment("used to store references to external systems, e.g. Stripe").
+			Optional(),
+		field.String("external_subscription_id").
+			Comment("used to store references to external systems, e.g. Stripe").
+			Optional(),
+		field.Bool("expires").
+			Comment("whether or not the customers entitlement expires - expires_at will show the time").
+			Default(false),
```
golanglemonade, line 32: Should this be `expired` ?  Or is this "will it expire in the future?

internal/ent/schema/entitlement.go
```diff
@@ -18,15 +18,27 @@ type Entitlement struct {
...
+		field.Time("expires_at").
+			Comment("the time at which a customer's entitlement will expire, e.g. they've cancelled but paid through the end of the month").
+			Optional().
+			Nillable(),
+		field.Bool("cancelled").
+			Comment("whether or not the customer has cancelled their entitlement - usually used in conjunction with expires and expires at").
+			Default(false),
+		field.String("organization_id").
+			Comment("the ID of the organization associated with the entitlement"),
```
golanglemonade, line 41: This should already be covered by the `owner` edge?

internal/ent/schema/entitlement.go
matoszz: "does it expire in the future" true / false

internal/ent/schema/group.go
```diff
@@ -55,6 +64,9 @@ func (Group) Fields() []ent.Field {
 			Annotations(
 				entgql.OrderField("display_name"),
 			),
+		field.String("organization_id").
+			Comment("the ID of the organization the group belongs to").
+			Optional(),
```
golanglemonade, line 69: Same here, covered by the `owner` edge?

golanglemonade (approved):

PR #273, opened 2023-12-19
add authz group tests

golanglemonade:


internal/ent/privacy/rule/group.go
```diff
@@ -18,29 +18,31 @@ func HasGroupReadAccess() privacy.GroupQueryRuleFunc {
...
-		}
-
-		if access {
-			q.Logger.Infow("access allowed", "relation", fga.CanView, "group_id", gID)
-
-			return privacy.Allow
+		if gCtx != nil {
+			// check group id from graphql arg context
+			// when all groups are requested, the interceptor will check group access
+			gID, ok := gCtx.Args["id"].(string)
```
matoszz, line 24: do you actually want "id" or "owner"?

internal/graphapi/group_test.go
```diff
@@ -2,18 +2,109 @@ package graphapi_test
...
+
+			require.NoError(t, err)
+			require.NotNil(t, resp)
+			require.NotNil(t, resp.Group)
+		})
+	}
+
+	// delete created org and group
+	(&OrganizationCleanup{OrgID: org1.ID}).MustDelete(reqCtx)
+	(&GroupCleanup{GroupID: group1.ID}).MustDelete(reqCtx)
```
matoszz, line 105: does ordering here matter? should you not be able to delete an org with resources in it and/or should it not cascade? I'd have assumed you needed to at least delete the group before the organization

internal/graphapi/group_test.go
```diff
@@ -172,6 +456,128 @@ func TestQuery_GroupNoAuth(t *testing.T) {
...
+	owner2 := (&OrganizationBuilder{}).MustNew(reqCtx)
+
+	listObjects := []string{fmt.Sprintf("organization:%s", owner1.ID)}
+
+	testCases := []struct {
+		name        string
+		groupName   string
+		description string
+		displayName string
+		logoURL     string
```
matoszz, line 497: I'd suggest leaving optional fields not relevant to group creation or modification out of the tests as we're likely to rename, shuffle around, and having tests break for optional fields is annoying

internal/ent/interceptors/group.go
```diff
@@ -0,0 +1,78 @@
...
+				return filterGroupsByAccess(ctx, q, v)
+			}
+
+			return next.Query(ctx, q)
+		})
+	})
+}
+
+// filterGroupsByAccess checks fga, using ListObjects, and ensure user has view access to a group before it is returned
+func filterGroupsByAccess(ctx context.Context, q *generated.GroupQuery, v ent.Value) ([]*generated.Group, error) {
```
matoszz, line 36: should this also not enumerate over organizations given you should be able to view groups within an org if you have view access to the org? (i don't believe we create tuples unless you created the group or are a member of the group vs. being an organization member)

internal/ent/privacy/rule/group.go
golanglemonade: id, because I'm checking the group. The org owner parts are already checked because of how privacy checks happen at the schema level, when you get an organization as part of the query it will check the access to the org as well.

internal/graphapi/group_test.go
golanglemonade: I actually testing this and switched them to see because I thought it would fail but it didn't and I think this is because this actually deletes, but the cascade soft deletes only soft deletes. I'm not entirely sure though. I switch them just in case it breaks things later though

internal/graphapi/group_test.go
golanglemonade: I'm okay with doing that, assuming we go back later and add them. I like having the tests so that I _know_ when I break something, and consciously remove the test

internal/ent/interceptors/group.go
golanglemonade: Thats covered by the org checks because these checks happens on the schema level. If you look at the tests, you'll see calls to `listObjects` happens for both orgs + groups on the group tests because of that.

matoszz (approved):

Issue #136, opened 2023-11-26
Rollback transaction if writes to FGA fail


golanglemonade:
Tuple writes (creates and deletes) happen using an ent `Hook` which allows this to happen during the same flow as the object creation/deletion in the database. 

If the tuple write fails due to any reason, we should rollback the transaction: https://github.com/datumforge/datum/blob/main/internal/ent/hooks/organization.go#L57

Look at transactions on Hooks: https://entgo.io/docs/transactions/#hooks to achieve this

PR #283, opened 2023-12-20
switch mutations to transacationer client from context

golanglemonade:
Previously, when an error occurred during a mutation, we could be left with orphaned resources. For example, if in the edge cleanup of an organization, a cascade delete failed, the organization would be deleted but the child resources would not be. Now, with the transactioner client, if an error occurs, nothing is committed to the database. 

This was tested by causing an error to be returned in the edge cleanup, when an error was returned: 

```
go run cmd/cli/main.go org delete -i 01HJ494EXYJ0M9MNQDEJM6TD9G                         
Error: {"networkErrors":null,"graphqlErrors":[{"message":"error deleting associated objects: ent/privacy: deny rule","path":["deleteOrganization"]}]}
```

The organization was not soft deleted: 
```
 go run cmd/cli/main.go org get -i 01HJ494EXYJ0M9MNQDEJM6TD9G
{
  "organization": {
    "children": {},
    "createdAt": "2023-12-20T11:52:14.142068-07:00",
    "createdBy": "01HHFK49TSES1PMH34SXG6TEH2",
    "displayName": "Transaction Test 3",
    "id": "01HJ494EXYJ0M9MNQDEJM6TD9G",
```

On `main` branch, doing the same test, the organization was deleted even though an error was returned. 

When no errors occur, organization is successfully deleted. 

This PR also removes federation settings that are not currently being used. 

Work for https://github.com/datumforge/datum/issues/136
Resolves https://github.com/datumforge/datum/issues/168


matoszz (approved):
Nice work @golanglemonade !

PR #288, opened 2023-12-21
add secure, cachecontrol, middleware

matoszz:
- adds new secure (XSS protection) middleware
- adds new Ratelimiter middleware
- adds new cache control headers and entity header removal
- adds new http bodydump middleware for printing out request / response payloads; this middleware has been added to several existing routes which now prints out both the request and response payloads
- registers the above middleware
- added several of the middlewares to existing routes to test and confirm - currently the ratelimiter is on an innocuous endpoint and you can run a few queries in succession to demonstrate it's working

Several of these middlewares need to have their configuration options pulled out as they are statically assigned right now and we may want to do things like assign different rate limits to different endpoints, protect some and not others, etc. This PR was mostly an effort to get these functional + added + tested and confirm they work with the rest of our components (especially adding the security checks to our login process so that we don't continue to add authentication / authorizations without the security checks in place)


internal/httpserve/middleware/debug/bodydump.go
```diff
@@ -0,0 +1,103 @@
...
+	"net/http"
+	"time"
+
+	echo "github.com/datumforge/echox"
+	"github.com/datumforge/echox/middleware"
+	"go.uber.org/zap"
+)
+
+// BodyDump prints out the request body for debugging purpose but attempts to obfuscate sensitive fields within the requests
+func BodyDump(l *zap.Logger) echo.MiddlewareFunc {
```
golanglemonade, line 15: It looks like you are doing `.Sugar()`  on all your logs, why not just pass in the suggared logger here?

internal/httpserve/middleware/ratelimit/ratelimiter.go
```diff
@@ -0,0 +1,52 @@
...
+		//		Skipper: func(c echo.Context) bool {
+		//			for _, path := range protectedPaths[c.Request().Method] {
+		//				if c.Path() == path {
+		//					return false
+		//				}
+		//			}
+		//
+		//			return true
+		//		},
+		Store: middleware.NewRateLimiterMemoryStoreWithConfig(
```
golanglemonade, line 24: Should this be deleted?

internal/httpserve/middleware/ratelimit/ratelimiter.go
```diff
@@ -0,0 +1,52 @@
...
+			middleware.RateLimiterMemoryStoreConfig{
+				Rate:      1,               // nolint: gomnd
+				Burst:     3,               // nolint: gomnd
+				ExpiresIn: 1 * time.Minute, // nolint: gomnd
+			},
```
golanglemonade, line 29: Does this mean 1 request per minute?

internal/httpserve/middleware/secure/secure.go
```diff
@@ -0,0 +1,28 @@
...
+	secureConfig := middleware.SecureConfig{
+		XSSProtection:      "1; mode=block",
+		ContentTypeNosniff: "nosniff",
+		XFrameOptions:      "SAMEORIGIN",
+		//		HSTSMaxAge:            31536000, // nolint: gomnd
+		//		HSTSExcludeSubdomains: true,
+		//		HSTSPreloadEnabled:    true,
+		ContentSecurityPolicy: "default-src 'self'",
+		Skipper: func(c echo.Context) bool {
+			return strings.Contains(c.Path(), "/api-docs")
```
golanglemonade, line 22: This isn't an endpoint we have?

internal/httpserve/middleware/debug/bodydump.go
```diff
@@ -0,0 +1,103 @@
...
+				}
+				// reset to return to the normal terminal color variables (kinda default)
+				resetColor = reset
+
+				name := "RESPONSE"
+				fmt.Printf("\n[%s] %v |%s %3d %s| %s\n",
+					name, // response
+					time.Now().Format("2006/01/02 - 15:04:05"), // TIMESTAMP for route access
+					statusColor, resStatus, resetColor, // response status
+					res.Header(),
```
golanglemonade, line 97: This will print all the cookies into the logs, which includes the access + refresh tokens - we should be careful about that

internal/httpserve/server/server.go
```diff
@@ -64,8 +67,12 @@ func (s *Server) StartEchoServer() error {
...
 		echozap.ZapLogger(s.logger),                  // add zap logger
 		echocontext.EchoContextToContextMiddleware(), // adds echo context to parent
-		cors.New(), // add cors middleware
-		mime.New(), // add mime middleware
+		cors.New(),                   // add cors middleware
+		mime.New(),                   // add mime middleware
+		cachecontrol.New(),           // add cache control middleware
+		ratelimit.RateLimiter(),      // add ratelimit middleware
+		middleware.Secure(),          // add XSS middleware
+		echodebug.BodyDump(s.logger), // adds debugging body dump middleware
```
golanglemonade, line 75: We only want to add this middleware when we have debug turned on (`srv.Debug`)

internal/httpserve/route/login.go
```diff
@@ -20,14 +25,16 @@ import (
 // without the user having to log in again. The refresh token overlaps with the access
 // token to provide a seamless authentication experience and the user can refresh their
 // access token so long as the refresh token is valid.
+var logger *zap.Logger
```
golanglemonade, line 28: this will give you a null logger, you need to pass the logger through - so I was trying to figure out why you weren't getting null pointer errors, and its because its actually using the middleware you set in `internal/httpserve/server/server.go` meaning you don't need to add these here (or in the other REST routes). Recover is the only one I've found so far that needed to be added in the REST handlers specifically that were added on the global level.

internal/httpserve/route/login.go
```diff
@@ -20,14 +25,16 @@ import (
...
+
 func registerLoginHandler(router *echo.Echo, h *handlers.Handler) (err error) {
 	_, err = router.AddRoute(echo.Route{
 		Method: http.MethodPost,
 		Path:   "/login",
 		Handler: func(c echo.Context) error {
 			return h.LoginHandler(c)
 		},
-		Middlewares: []echo.MiddlewareFunc{middleware.Recover()},
+		Middlewares: []echo.MiddlewareFunc{middleware.Recover(), dump.BodyDump(logger), secure.Secure(), ratelimit.RateLimiter(), cachecontrol.New()},
```
golanglemonade, line 37:
```suggestion
		Middlewares: []echo.MiddlewareFunc{middleware.Recover()},
```

See comment above, you should be bale to remove all these from the handler registration

internal/httpserve/route/login.go
golanglemonade, line 28:
```suggestion
```

internal/httpserve/middleware/ratelimit/ratelimiter.go
matoszz: Yea, I kept it intentionally low so you could see it while testing locally

internal/httpserve/middleware/secure/secure.go
matoszz: it will be in a PR I already have staged I was just trying to cut down on the amount of changes in this one

internal/httpserve/middleware/debug/bodydump.go
matoszz: yea that was intentional ;) and yes, this is only for testing only. I can make a TODO to roll this under only when the server is started in development mode or something

internal/httpserve/middleware/ratelimit/ratelimiter.go
matoszz: i mean, maybe - it's the way I wanted to implement it but couldnt figure out how to make it work

internal/httpserve/middleware/debug/bodydump.go
matoszz: because zap is dumb

internal/httpserve/server/server.go
matoszz: says YOU

matoszz: fixed :)

golanglemonade (approved):

PR #296, opened 2023-12-22
Add cascade deletes to user edges

tylerauerbeck:
- Adds builder test function for personalaccesstoken
- Adds cascade deletes to user schema for personalaccesstoken, sessions and usersetting
- Adds test for user cascade deletes


internal/graphapi/models_test.go
```diff
@@ -127,3 +137,32 @@ func (g *GroupCleanup) MustDelete(ctx context.Context) {
...
+
+	return EntClient.PersonalAccessToken.Create().
+		SetName(t.Name).
+		SetOwnerID(t.OwnerID).
+		SetOwner(owner).
+		SetToken(t.Token).
+		SetDescription(t.Description).
+		SetExpiresAt(t.ExpiresAt).
+		SaveX(ctx)
+}
```
golanglemonade, line 168: If t.OwnerID is set, we shouldn't need to create a new user as well? Do we need to set both `SetOwnerID` and `SetOwner` in the Create? 

```suggestion
	if t.OwnerID == "" {
		owner := (&UserBuilder{}).MustNew(ctx)
		t.OwnerID = owner.ID
	} 

	return EntClient.PersonalAccessToken.Create().
		SetName(t.Name).
		SetOwnerID(t.OwnerID).
		SetToken(t.Token).
		SetDescription(t.Description).
		SetExpiresAt(t.ExpiresAt).
		SaveX(ctx)
}
```

golanglemonade (approved):
nice @tylerauerbeck !

golanglemonade (approved):

PR #310, opened 2023-12-23
create dev yaml config for auth'ed server

golanglemonade:
- Creates a dev config to replace cli flags for the `run-dev-auth` setup
- Fixes issue with parsing durations from the config


matoszz (approved):
ahhh you're the best! thanks for getting this started!

PR #312, opened 2023-12-24
be more consistent with the logger, use sugared

golanglemonade:
the echozap middleware currently uses the normal zap logger, but aligning everything else to always use the sugared logger.


matoszz (approved):
would rather have a different logger entirely but ok i guess if you want

PR #317, opened 2023-12-27
create email verification tokens schemas and generate

matoszz:
- adds email verification tokens and edges to user
- email verification tokens schema will hold the tokens we generate to send to a user to verify their emails

The associated HTTP handlers / routes will be added in a later PR


internal/ent/schema/emailverificationtoken.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+// Fields of the EmailVerificationToken
+func (EmailVerificationToken) Fields() []ent.Field {
+	return []ent.Field{
+		field.String("token").
+			Comment("the verification token sent to the user via email which should only be provided to the /verify endpoint + handler").
+			Unique(),
+		field.Time("ttl").
+			Comment("the ttl of the verification token which needs to be explicitly set, otherwise defaults to the current time").
```
golanglemonade, line 27:
```suggestion
			Comment("the ttl of the verification token which needs to be explicitly set").
```

This no longer sets a default, so this comment should be updated + schema regenerated.

internal/ent/schema/emailverificationtoken.go
```diff
@@ -0,0 +1,70 @@
...
+
+// EmailVerificationToken holds the schema definition for the EmailVerificationToken entity
+type EmailVerificationToken struct {
+	ent.Schema
+}
+
+// Fields of the EmailVerificationToken
+func (EmailVerificationToken) Fields() []ent.Field {
+	return []ent.Field{
+		field.String("token").
```
golanglemonade, line 23: This should be `NotEmpty().` ?

internal/ent/schema/emailverificationtoken.go
```diff
@@ -0,0 +1,70 @@
...
+			Comment("the verification token sent to the user via email which should only be provided to the /verify endpoint + handler").
+			Unique(),
+		field.Time("ttl").
+			Comment("the ttl of the verification token which needs to be explicitly set, otherwise defaults to the current time").
+			Nillable(),
+		field.String("email").
```
golanglemonade, line 29: Same here, `NotEmpty().`?

golanglemonade (dismissed):

internal/ent/schema/emailverificationtoken.go
matoszz: its a required field; does NotEmpty() actually signify something on a required string field?

golanglemonade (approved):

PR #329, opened 2023-12-28
Split auth override into sep. compose file

tylerauerbeck:
This splits the auth overrides for the API into a separate compose file so that the `fga` compose stack can still be run independently.

Now the compose stack is:
- docker-compose.yml - API with no auth
- docker-compose-auth.yml - API with auth configured (but without FGA components)
- docker-compose-fga.yml - FGA and Postgres

Which can be used with the following tasks:
`task run-dev-auth` - spins up FGA compose stack, but runs API locally
`task compose:datum` - spins up compose stack with just the API (no auth, no fga)
`task compose:datum-auth` - spins up compose stack with Auth AND FGA
`task compose:fga` - spins up fga compose stack, no API

### DIFF
--- a/Taskfile.yaml
+++ b/Taskfile.yaml
@@ -166,7 +166,7 @@ tasks:
   compose:datum-auth:
     desc: brings up the compose environment for the datum server configured with auth
     cmds:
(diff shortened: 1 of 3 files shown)


golanglemonade (approved):
dope, thanks @tylerauerbeck !

PR #332, opened 2023-12-28
Update graph generation to entc; Drop gen_graph.sh

### LINKED ISSUE #108: Test / validate if `entc` templates can be used to replace `/scripts/gen_graphql.sh`, otherwise make bash smarter
Today we run into issues when running the `gen_graphql` script due to things like camel case variables; the script also does not attempt to detect the presence of annotations added to fields / schemas and should be smarter + more integrated to ensure when annotations are used for exclusion(s) in schema files those also flow down to our graphql interfaces

tylerauerbeck:
This updates the graphql generation to be done via `entc` and drops the need for `gen_graph.sh`

This could probably be made a bit more elegant, but it uses the same logic from gen_graph.sh to skip specified files, loop through `internal/ent/schema` files and then generate the `*.graphql` files found in `/schemas/*.graphql` at the root of the repo. I've updated this generator to be a bit smarter by grabbing the Schema name from the struct in each file under `internal/ent/schema` so that we have the right casing for each object instead of having to piece it together ourselves.

Should begin to address / resolve #108


internal/ent/entc.go
```diff
@@ -88,3 +98,63 @@ func main() {
...
 	}
 }
+
+func generateSchemaFuncs() {
+	r, err := regexp.Compile("type ([a-zA-Z].*) struct")
+	if err != nil {
+		log.Fatalf("Unable to compile regex: %v", err)
+	}
+
+	skipFile, err := os.ReadFile("./scripts/files_to_skip.txt")
```
matoszz, line 108: nice

matoszz (commented):
@tylerauerbeck mind updating the task file references to gen_schema.sh as well so it's called appropriately

golanglemonade (approved):

PR #344, opened 2023-12-30
add transaction middelware for rest endpoints

golanglemonade:
this removes the need to manually remember to commit/rollback transactions on request success failure by used a middleware. The transaction client is added to the context so it can be used by all queries and mutations. This should look the same as the existing Transactioner middleware within the graph endpoints.

When something fails in the request, you will see the following in the debug logs:

```
{"level":"debug","ts":1703973401.2261078,"caller":"transaction/transaction.go:59","msg":"rolling back transaction in middleware","app":"datum"}
{"level":"debug","ts":1703973401.226114,"logger":"ent","caller":"dialect/dialect.go:79","msg":"Tx(676f64a7-c6f6-43d4-9b2b-2f23a1ce6fcd): rollbacked","app":"datum"}
```

### DIFF
--- a/internal/httpserve/route/routes.go
+++ b/internal/httpserve/route/routes.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/datumforge/datum/internal/httpserve/handlers"
 	"github.com/datumforge/datum/internal/httpserve/middleware/ratelimit"
+	"github.com/datumforge/datum/internal/httpserve/middleware/transaction"
 )
 
 const (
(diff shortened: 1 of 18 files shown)


internal/httpserve/route/routes.go
```diff
@@ -27,6 +28,14 @@ type Route struct {
 
 // RegisterRoutes with the echo routers
 func RegisterRoutes(router *echo.Echo, h *handlers.Handler) error {
+	// add transaction middleware
+	transactionConfig := transaction.Client{
```
matoszz, line 32: great idea @golanglemonade <3

matoszz (approved):

matoszz (approved):

PR #346, opened 2023-12-31
Adds session middleware

matoszz:
Adds a session token issuing middleware


internal/httpserve/handlers/register_test.go
```diff
@@ -92,9 +93,12 @@ func TestRegisterHandler(t *testing.T) {
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
 			// create echo context with middleware
-			e := setupEcho()
+			e := setupEcho(h.SM)
+
 			e.POST("register", h.RegisterHandler)
 
+			e.Use(session.LoadAndSave(h.SM))
+
```
golanglemonade, line 101:
```suggestion
```

You shouldn't need this since its setup in `e := setupEcho(h.SM)`

internal/httpserve/middleware/session/session.go
```diff
@@ -0,0 +1,114 @@
...
+
+			var token string
+
+			cookie, err := c.Cookie(config.SessionManager.Cookie.Name)
+
+			if err == nil {
+				token = cookie.Value
+			}
+
+			ctx, err = config.SessionManager.Load(ctx, token)
```
golanglemonade, line 60: token will be an empty string here if `c.Cookie(config.SessionManager.Cookie.Name)` returned an error. Instead should the above be changed to: 

```
			cookie, err := c.Cookie(config.SessionManager.Cookie.Name)
			if err != nil {
				return err
			}
  
                         token = cookie.Value
```

golanglemonade (approved):

Issue #331, opened 2023-12-28
Sever config should use envconfig instead of viper flags


golanglemonade:
We have some things using envconfig, and some using viper. All configs for the server (not the cli) should switch to using envconfig https://github.com/kelseyhightower/envconfig

- Switch to this
- Update yaml config to use this/env vars

PR #349, opened 2023-12-31
switch to envconfig from viper flags for server settings

golanglemonade:
All in-use env vars were added to `config/.env.example` and the dev + dev-auth settings have their own `.env` files. I believe I found all the places where we were previously using flags, and they have been switched to use the env vars. 

The only two flags left `--debug` and `--pretty` are for logging purposes and are needed earlier in the process so they will remain viper flags. 

The auth-provider settings were not added to .env.example because they are not currently in use, a TODO was added to do this once we begin using those settings.

Resolves https://github.com/datumforge/datum/issues/331


matoszz (approved):
such a great refactor; will help a ton with deployments! <3

Issue #326, opened 2023-12-28
Publish all-in-one-image


tylerauerbeck:
There are use cases that would be better supported by having all dependencies wrapped up into a single image. To support this, an an all-in-one image should be published that contains all necessary components required to run the stack:
- API
- FGA
- Postgres(?)

PR #350, opened 2023-12-31
Initial all-in-one image

tylerauerbeck:
This adds the necessary Dockerfile to build an image that contains the following:
- OpenFGA running with in memory datastore
- Datum API connected to local FGA instance

Adds the following task to build this image:
`task docker:build:aio`

Which will build an image called `datum:dev-aio`

I haven't added this to a pipeline at the moment as I wasn't sure if this is something to be made publicly available or if we can expect this to be more of a dev-time utility

### DIFF
--- a/Taskfile.yaml
+++ b/Taskfile.yaml
@@ -159,7 +159,11 @@ tasks:
   docker:build:
     desc: builds the datum docker image
     cmds:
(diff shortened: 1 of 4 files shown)


golanglemonade (approved):

golanglemonade:
Resolves https://github.com/datumforge/datum/issues/326

PR #352, opened 2023-12-31
Ent Privacy rules for Users, Tokens

matoszz:
- Creates wrappers for generating new contexts based on tokens provided in original context
- Adds a "user owned" schema in attempt to wrap privacy decisions based on whether the object is owned by a user or not
- Creates ent privacy rules / decisions on mutations and queries

This currently removes the users returned from groups queries. This will be added back in a subsequent PR that adds users to groups.


matoszz (commented):
LGTM!

golanglemonade (approved):

golanglemonade (approved):

PR #354, opened 2024-01-02
Add initial support for postgres supported databases

### LINKED ISSUE #115: Add drivers + support + tests for PostgreSQL (and cockroachdb?)
We chose the Atlas tool chain for database portability and are focused primarily on vetting a usecase with Turso + SQLLite for our product MVP, but should build in full support for PostgreSQL (and possibly CockroachDB) given a few factors:

- We aren't yet fully certain we can / should use Turso, so should have a production-ready DB compatibility + support as a fallback
- Open source consumers of the software should not be married to our production runtime decisions (e.g. Turso) and would also desire a production-ready database to make the stack viable for use
- Other required components to run the stack (e.g. openFGA) require postgresql and we (and OSS community) may desire a more consolidated database deployment scheme

tylerauerbeck:
This adds initial support for postgres'ish databases. This will support pointing the API against any database that supports the postgres protocol by setting the following in your `env` file:

```sh
DATUM_DB_DRIVE_NAME=postgres
DATUM_DB_PRIMARY_DB_SOURCE=postgres://$USERNAME:$PASSWORD@$HOST:$PORT/$DBNAME
```

Just updating the tests to be able to run against both sqlite and postgres databases and then this should be good. Has been manually tested against postgres to test initial functionality

resolves #115


tylerauerbeck:
```sh
datum=# \dt
                   List of relations
 Schema |           Name            | Type  |  Owner
--------+---------------------------+-------+----------
 public | email_verification_tokens | table | postgres
 public | entitlements              | table | postgres
 public | group_settings            | table | postgres
 public | group_users               | table | postgres
 public | groups                    | table | postgres
 public | integrations              | table | postgres
 public | oauth_providers           | table | postgres
 public | oh_auth_too_tokens        | table | postgres
 public | organization_settings     | table | postgres
 public | organizations             | table | postgres
 public | password_reset_tokens     | table | postgres
 public | personal_access_tokens    | table | postgres
 public | sessions                  | table | postgres
 public | user_organizations        | table | postgres
 public | user_settings             | table | postgres
 public | users                     | table | postgres
(16 rows)
```

.buildkite/pipeline.yaml
```diff
@@ -27,8 +27,14 @@ steps:
           command: ["golangci-lint", "run", "-v", "--timeout", "10m"]
           environment: 
             - "GOTOOLCHAIN=auto"
-  - label: ":golang: go test"
+  - label: ":golang: go test - {{matrix}}"
     key: "go_test"
+    env:
+      TEST_DB_URL: "{{matrix}}"
+    matrix:
```
golanglemonade, line 34: 😍

sonar-project.properties
```diff
@@ -5,11 +5,12 @@ sonar.projectName=datum
...
-sonar.exclusions=**/*_test.go,**/vendor/**,internal/ent/generated/**,internal/datumclient/**,query/**,schema/**,scripts/**,gen_schema.go,generate.go,tools.go,docker/**,db/**,internal/graphapi/gen_server.go,internal/ent/entc.go
+
+sonar.exclusions=**/*_test.go,**/vendor/**,internal/ent/generated/**,internal/datumclient/**,query/**,schema/**,scripts/**,gen_schema.go,generate.go,tools.go,docker/**,db/**,internal/graphapi/gen_server.go,internal/ent/entc.go,internal/testutils/**
 sonar.tests=.
 sonar.test.inclusions=**/*_test.go
 sonar.test.exclusions=**/vendor/**
 
 sonar.sourceEncoding=UTF-8
-sonar.go.coverage.reportPaths=coverage.out
+sonar.go.coverage.reportPaths=coverage.out-sqlite*
```
matoszz, line 15: is there a reason there's not a postgreSQL coverage out? not really a deal breaker but just curious

.buildkite/pipeline.yaml
```diff
@@ -27,13 +27,27 @@ steps:
           command: ["golangci-lint", "run", "-v", "--timeout", "10m"]
           environment: 
             - "GOTOOLCHAIN=auto"
-  - label: ":golang: go test"
-    key: "go_test"
+  - label: ":golang: go test - sqlite"
```
golanglemonade, line 30: Why can't sqlite be part of the matrix test?

tylerauerbeck: So.. this is kind of me just giving up at the moment because if I put it all in one matrix, I need a way to differentiate between which `coverage.out` we want to consume. I tried using `{{matrix}}` as part of the `go test` command, but that ends up choking on the special characters in the command. So I don't have a _good_ way around this at the moment. So I was going to create an issue and make this a future me problem.

matoszz (approved):
thanks for the hard work adding this @tylerauerbeck !

PR #359, opened 2024-01-03
Add Password Reset Handler, Routes, token validation

matoszz:
- adds pw reset httpserve/ handler, routes
- adds datumclient input for password reset
- adds getUserByResetToken and updateUserPassword to ent.go helpers


internal/datumclient/reset.go
```diff
@@ -0,0 +1,61 @@
...
+	"io"
+	"net/http"
+	"net/url"
+
+	"github.com/datumforge/datum/internal/httpserve/handlers"
+	"github.com/datumforge/datum/internal/httpserve/route"
+)
+
+// Reset a user password
+func Reset(c *Client, ctx context.Context, r handlers.RegisterRequest) (*handlers.ResetPasswordReply, error) {
```
golanglemonade, line 17:
```suggestion
func Reset(c *Client, ctx context.Context, r handlers. ResetPasswordRequest) (*handlers.ResetPasswordReply, error) {
```

internal/httpserve/handlers/ent.go
```diff
@@ -186,3 +204,21 @@ func (h *Handler) setEmailConfirmed(ctx context.Context, user *ent.User) error {
...
 }
+
+// updateUserPassword changes a user's password to a new dk hash value based on input
+func (h *Handler) updateUserPassword(ctx context.Context, id string, password string) error {
+	hash, err := passwd.CreateDerivedKey(password)
+	if err != nil {
+		return err
+	}
+
+	if _, err := transaction.FromContext(ctx).User.Update().SetPassword(hash).
```
golanglemonade, line 215: I discovered yesterday there is a `UpdateOneID(id)` you can use instead of `Update()` which allows you to skip the `Where` here

internal/httpserve/handlers/resetpassword.go
```diff
@@ -0,0 +1,126 @@
...
+
+	if token.ExpiresAt, err = user.GetPasswordResetExpires(); err != nil {
+		h.Logger.Errorw("unable to parse expiration", "error", err)
+
+		return ctx.JSON(http.StatusInternalServerError, ErrUnableToVerifyEmail)
+	}
+
+	// Verify the token with the stored secret
+	if err = token.Verify(user.GetPasswordResetToken(), user.PasswordResetSecret); err != nil {
+		return err
```
golanglemonade, line 87: These are responses to the user, we should log the error and return a json response

internal/datumclient/reset.go
```diff
@@ -0,0 +1,61 @@
...
+	out := handlers.ResetPasswordReply{
+		Message: "password reset successfully",
+	}
+
+	if err := json.NewDecoder(resp.Body).Decode(&out); err != nil {
+		return nil, err
+	}
+
+	if resp.StatusCode != http.StatusCreated {
+		return nil, newRegistrationError(resp.StatusCode, out.Message)
```
golanglemonade, line 57: Message is set to `password reset successfully` - you probably don't want that in a failure mode though

golanglemonade (approved):

internal/httpserve/handlers/ent.go
```diff
@@ -181,6 +202,19 @@ func (h *Handler) setEmailConfirmed(ctx context.Context, user *ent.User) error {
...
+		return err
+	}
+
+	return nil
+}
+
+// updateUserPassword changes a updates a user's password in the database
+func (h *Handler) updateUserPassword(ctx context.Context, id string, password string) error {
+	if _, err := transaction.FromContext(ctx).User.UpdateOneID(id).SetPassword(password).Save(ctx); err != nil {
+		h.Logger.Errorw("error updating user password", "error", err)
```
matoszz, line 216: would this potentially return the clear-text input `password` value in the error?

internal/httpserve/handlers/resetpassword.go
```diff
@@ -0,0 +1,193 @@
...
+	if token.ExpiresAt, err = user.GetPasswordResetExpires(); err != nil {
+		h.Logger.Errorw("unable to parse expiration", "error", err)
+
+		return ctx.JSON(http.StatusInternalServerError, ErrUnableToVerifyEmail)
+	}
+
+	// Verify the token with the stored secret
+	if err = token.Verify(user.GetPasswordResetToken(), user.PasswordResetSecret); err != nil {
+		if errors.Is(err, tokens.ErrTokenExpired) {
+			_, err := h.storeAndSendPasswordResetToken(ctx.Request().Context(), user)
```
matoszz, line 111: in the event that the password-reset token is expired, I don't think we should automatically issue another one; the user should be required to re-request it

internal/httpserve/handlers/resetpassword.go
```diff
@@ -0,0 +1,193 @@
...
+		return ctx.JSON(http.StatusBadRequest, ErrorResponse(err))
+	}
+
+	// make sure its not the same password as current
+	valid, err := passwd.VerifyDerivedKey(*entUser.Password, rp.Password)
+	if err != nil || valid {
+		return ctx.JSON(http.StatusBadRequest, ErrorResponse(auth.ErrNonUniquePassword))
+	}
+
+	if err := h.updateUserPassword(ctx.Request().Context(), entUser.ID, rp.Password); err != nil {
```
matoszz, line 134: same question as in another block; are we returning the input string (password) in the error response?

golanglemonade: no, the password (or any input ) is returned the response.

internal/httpserve/handlers/ent.go
golanglemonade: no, and this log does not log the input (password or otherwise)

golanglemonade (approved):

PR #363, opened 2024-01-04
Tyoescript codegen

matoszz:
Not entirely sure if the generated output is valuable but figured it was worthwhile to get a configuration going and have our resident UI expert(s) take a look :)


golanglemonade:
This was added to the `datum-ui` repo for now

PR #376, opened 2024-01-07
Add schemagen hook

tylerauerbeck:
This generates our schema `.graphql` based off of a codegen hook. This allows us to more easily get at the things we need inside of our templates. It also allows us to use a schema annotation to indicate what schemas should skip codegen instead of using a static file with filenames. 

During this cleanup, it was noticed that some schemas weren't following the same format as the majority of others. Cleanup was done during this PR to bring these inline so that we don't run into any unexpected UX later down the road.


golanglemonade (approved):

tylerauerbeck:
Not sure why this looks like it's failed, but the build is happy: https://buildkite.com/datum/datum/builds/1527 

I think this may just be a remnant of a branch I pushed directly earlier instead of running from my fork.

golanglemonade (approved):

golanglemonade (approved):

PR #381, opened 2024-01-07
Add initial support for otel tracing

tylerauerbeck:
This adds initial support for otel tracing. This adds the default tracing provided by `otelsql` and `otelgqlgen`. This gives us visibility to things like this:

Screenshot 2024-01-07 at 12 51 06 PM

Additional custom attributes/traces can be added as necessary in follow up PRs.

This also provides a compose environment to spin up a collector and jaeger UI to be able to view these traces. By default the `.env` example files configure tracing to just go to stdout, but can be updated to use the `otelhttp` provider to send traces to the collector so they are visible in the jaeger UI.


docker/docker-compose-tracing.yml
```diff
@@ -0,0 +1,34 @@
...
+    restart: always
+    ports:
+      - "6831:6831/udp"
+      - "6832:6832/udp"
+      - "5778:5778"
+      - "16686:16686"
+      - "14250:14250"
+      - "14268:14268"
+      - "14269:14269"
+      - "9411:9411"
```
golanglemonade, line 16: Can you add comments for what these are for like you did on the otel-collector

golanglemonade (approved):

docker/docker-compose-tracing.yml
tylerauerbeck: Yep. I'm actually just going to cut this down to just expose the UI. The other communication(s) will happen within the docker network so we don't need to expose them.

golanglemonade (approved):

Issue #150, opened 2023-11-28
Add tests ensuring user personal orgs cleaned up


matoszz:
Add tests confirming that the personal organization tied to a user is deleted / cleaned up when the user is deleted

related: https://github.com/datumforge/datum/pull/146

Issue #377, opened 2024-01-07
Investigate flakey user tests


tylerauerbeck:
We've seen a handful of occasions (on different tests with the `TestMutation_CreateUser` test where the test(s) will fail with `personal org expected to be created`. We should investigate what's causing these tests to flake and get that addressed. See example below:

```
FAIL: TestMutation_CreateUser (0.08s)
    --- FAIL: TestMutation_CreateUser/Create_no_display_name,_should_default_to_email (0.01s)
        user_test.go:261:
            	Error Trace:	/workdir/internal/graphapi/user_test.go:261
            	Error:      	Should be true
            	Test:       	TestMutation_CreateUser/Create_no_display_name,_should_default_to_email
            	Messages:   	personal org expected to be created
FAIL
```

golanglemonade:
I haven't seen this in a few days, when it used to fail frequently. I believe the test update in the linked PR fixed this.

Issue #293, opened 2023-12-22
Add logic to add users to organizations and groups


golanglemonade:
Right now, when a group or organization is created the user that made the request is given access, via authz tuples, to the organization, but the user is not actually added to the group as a user. 

1. Assign user of personal org to the org
2. Assign user that creates and org or group, to the org
4. ability to assign a user to a group or organization as the admin/owner

PR #382, opened 2024-01-07
add roles, group and org memberships, org owned mixin

golanglemonade:
- Adds group_membership and org_membership join tables
- Updates user hooks + tests to add user as admin on the created personal org
- Adds org owned mixin, and added to both groups and entitlements

Follow ups for another pr: 
- implement resolvers for group and org memberships
- FGA tuple creation based on roles 
- Add a roles schema to enable new roles to be created

Partially addresses https://github.com/datumforge/datum/issues/293
Resolves https://github.com/datumforge/datum/issues/150
Updates personal org test, which may address https://github.com/datumforge/datum/issues/377


internal/ent/enums/enums.go
```diff
@@ -0,0 +1,47 @@
...
+	"io"
+)
+
+type Role string
+
+const (
+	RoleOwner    Role = "OWNER"
+	RoleAdmin    Role = "ADMIN"
+	RoleMember   Role = "MEMBER"
+	RoleCustomer Role = "CUSTOMER"
```
matoszz, line 14: did you want to leave this in here?

matoszz (dismissed):

internal/ent/enums/enums.go
golanglemonade: nah, I'll it out for now since its not used

matoszz (approved):

PR #395, opened 2024-01-09
cascade delete through edges, soft delete org + group members

golanglemonade:
- Adds a new annotation to cascade delete through edges. This annotation has to be on the node, instead of the edge, because of how through edges work we will get duplicate code if the annotation is on the edge
- Adds soft delete mixin to org and group members


internal/ent/templates/edge_cleanup.tmpl
```diff
@@ -24,6 +24,18 @@
                         }
                     {{ end }}
                {{- end }}
+               {{- if $annotation := $node.Annotations.DATUM_CASCADE_THROUGH }}
+               {{- range $schema := $annotation.Schemas }}
+                    {{- $field := $schema.Through }}
+                    {{/* use the client to delete records where the edge has a field and a through schema (provided by the annotation) containing the ID provided by the func */}}
+                    if exists, err := FromContext(ctx).{{ $field }}.Query().Where(({{ $field | lower }}.Has{{ $schema.Field }}With({{ $node.Name | lower }}.ID(id)))).Exist(ctx); err == nil && exists {
```
matoszz, line 31: do these statements not need to be wrapped `{{ }}`?

internal/entx/annotation.go
```diff
@@ -2,13 +2,28 @@ package entx
 
 // CascadeAnnotationName is a name for our cascading delete annotation
 var CascadeAnnotationName = "DATUM_CASCADE"
+
+// CascadeThroughAnnotationName is a name for our cascading through edge delete annotation
+var CascadeThroughAnnotationName = "DATUM_CASCADE_THROUGH"
```
matoszz, line 7: should these be prefixed with Datum rather than a more logical prefix?

matoszz (approved):

internal/ent/templates/edge_cleanup.tmpl
golanglemonade: nope, this is similar to the loop through the edges, just on the node instead. The `if` here is what is actually added to the `edge_cleanup.go`, not the `if` on the conditional of the template, if that makes sense

PR #408, opened 2024-01-11
fix unique index with soft deletes

golanglemonade:
- fixes unique indices with soft deletes on users, group members, and org members tables
- makes `user_id` and `group/org_id` immutable on membership tables (meaning only the role can be updated)


internal/httpserve/handlers/register_test.go
```diff
@@ -55,16 +55,6 @@ func TestRegisterHandler(t *testing.T) {
 			emailExpected:  true,
 			expectedStatus: http.StatusCreated,
 		},
-		{
```
golanglemonade, line 58: This was a bad test, the user was deleted in each test therefore this was only passing because the index did not take into account the `deleted_at` field. A new test was added to `user_test` to cover this case

matoszz (approved):

Issue #414, opened 2024-01-11
Prevent adding members to personal orgs


golanglemonade:
Personal orgs should only be allowed to have 1 member, the owner of the personal org. Prevent adding any other users to personal orgs.

Issue #407, opened 2024-01-11
Updating user role to the same results in FGA error


golanglemonade:
```
2024-01-10T17:58:34.456-0700	ERROR	resolvers	graphapi/orgmembership.resolvers.go:106	failed to update org member role	{"app": "datum", "error": "Write validation error for POST Write with body {\"code\":\"cannot_allow_duplicate_tuples_in_one_request\",\"message\":\"duplicate tuple in write: user: 'user:01HKTMJB90RB9NEZ43YQGNXVAX', relation: 'admin', object: 'organization:01HKTJ2HK94RX1D99DRQTJP9BS'\"}\n with error code cannot_allow_duplicate_tuples_in_one_request error message: duplicate tuple in write: user: 'user:01HKTMJB90RB9NEZ43YQGNXVAX', relation: 'admin', object: 'organization:01HKTJ2HK94RX1D99DRQTJP9BS'"}
```

This happens on update of a role where old role == new role, because a transaction will be sent to FGA that will contain the same update and delete.

If oldRole == newRole, return success and do not making an update. 

Details:
```
go run cmd/cli/main.go orgmembers get -o 01HKX1XF8CXZBWSX2VYGVSQC17
{
  "orgMemberships": {
    "edges": [
      {
        "node": {
          "id": "01HKX1XF8E895BPFG9M0WVQZ49",
          "orgID": "01HKX1XF8CXZBWSX2VYGVSQC17",
          "role": "OWNER",
          "user": {
            "displayName": "sfunk1@datum.net",
            "email": "sfunk1@datum.net",
            "firstName": "Sarah",
            "id": "01HKWMV5X2HTSF010ZEB92V2PS",
            "lastName": "Funkhouser"
          },
          "userID": "01HKWMV5X2HTSF010ZEB92V2PS"
        }
      },
      {
        "node": {
          "id": "01HKX24T074G5CG72KHXJ5EXAB",
          "orgID": "01HKX1XF8CXZBWSX2VYGVSQC17",
          "role": "MEMBER",
          "user": {
            "displayName": "sfunk@datum.net",
            "email": "sfunk@datum.net",
            "firstName": "Sarah",
            "id": "01HKWMREC7CMVGG61E6A030W4S",
            "lastName": "Funkhouser"
          },
          "userID": "01HKWMREC7CMVGG61E6A030W4S"
        }
      }
    ]
  }
}
[...]

PR #418, opened 2024-01-12
add members as part of org update, org settings for create and update

golanglemonade:
- Adds ability to add members as part of org update
- Adds tests for adding/updating org settings as part of organization mutations
- Fixes issue with FGA transaction writes where when oldrole==newrole request would fail
- Prevents members from being added to personal orgs

Resolves: https://github.com/datumforge/datum/issues/407
Resolves: https://github.com/datumforge/datum/issues/414


internal/graphapi/organization_test.go
```diff
@@ -200,13 +201,16 @@ func TestMutation_CreateOrganization(t *testing.T) {
 	orgToDelete := (&OrganizationBuilder{}).MustNew(reqCtx)
 	// delete said org
 	(&OrganizationCleanup{OrgID: orgToDelete.ID}).MustDelete(reqCtx)
+	// setup user for membership
+	testUser = (&UserBuilder{}).MustNew(reqCtx)
```
golanglemonade, line 205:
```suggestion
```

matoszz (approved):

PR #432, opened 2024-01-16
Unleash the cookie monster

matoszz:
- removes unnecessary `domain` declarations
- pulls `cookie` out of the existing `store` location and into it's own section /internal/cookiemanager
- adds a few helper functions and default cookie + debug cookie for pre-loaded settings
- adds a few testing utilities and keygen functions I had tucked away for future things and easier to lump into this random PR 💯


internal/cookies/cookiemanager.go
```diff
@@ -0,0 +1,147 @@
...
+		cookie.Expires = expires
+	}
+
+	return cookie
+}
+
+// expiresTime converts a maxAge time in seconds to a time.Time in the future
+// ref http://golang.org/src/net/http/cookie.go?s=618:801#L23
+
+func expiresTime(maxAge int) (time.Time, bool) {
```
golanglemonade, line 83:
```suggestion
// ref http://golang.org/src/net/http/cookie.go?s=618:801#L23
func expiresTime(maxAge int) (time.Time, bool) {
```

internal/testutils/server.go
```diff
@@ -0,0 +1,52 @@
...
+
+	return client, server
+}
+
+// NewTestServerFunc is an adapter to allow the use of ordinary functions
+func NewTestServerFunc(handler func(w http.ResponseWriter, r *http.Request)) *httptest.Server {
+	return httptest.NewServer(http.HandlerFunc(handler))
+}
+
+// RewriteTransport rewrites https requests to http to avoid TLS tomfoolery
```
golanglemonade, line 39: We already have stuff like this in https://github.com/datumforge/datum/blob/main/internal/graphapi/tools_test.go#L248-L259 we should either consolidate or leave this out

matoszz: well forgive me for not seeing it tucked away in `graphapi` :P

internal/cookies/cookiemanager.go
```diff
@@ -0,0 +1,147 @@
+package cookies
+
+import (
+	"encoding/base64"
+	"net/http"
+	"time"
+)
+
+// CookieConfig configures http.Cookie creation
+type CookieConfig struct {
```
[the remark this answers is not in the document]
matoszz: that's the equivalent of saying we should duplicate code and configuration all over the place? it also prevents us from being able to set default configs and additionally leverage the package within other packages such as sessions to take advantage of securecookie functionality

golanglemonade (approved):

PR #472, opened 2024-01-31
Skipper configuration for auth and session middleware

golanglemonade:
- Creates a single session config with opts that is used by both the handlers and the graph route instead of each route create its own 
- Adds `Skipper` and `BeforeFunc` to the auth middleware
- Adds `WithOptions` to both auth and session middleware in order to configure `Skipper` and `BeforeFunc` options
- Moves `authmw` setup to `serveroptions` instead of `cmd/server`
- Uses the configuration middleware in `serveropts` instead of creating the arrays in `server.go`


matoszz (approved):
beautiful <3

PR #484, opened 2024-02-04
Posthog analytics

matoszz:
- adds new analytics package in internal
- init's the posthog client and creates a handler wrapper for various event creations
- adds the entc dependency required to use the analytics package within ent hooks
- adds basic events to org and org membership

### DIFF
--- a/internal/analytics/analytics.go
+++ b/internal/analytics/analytics.go
@@ -1,6 +1,8 @@
 package analytics
 
 import (
+	ph "github.com/posthog/posthog-go"
+
 	"github.com/datumforge/datum/internal/analytics/posthog"
 )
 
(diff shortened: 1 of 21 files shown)


internal/analytics/analytics.go
```diff
@@ -0,0 +1,101 @@
...
+func init() {
+	p := posthog.Init()
+	if p != nil {
+		handler = p
+	}
+}
+
+// EventManager isn't your normal party planner
+type EventManager struct {
+	EM Handler
```
golanglemonade, line 24: nit: 

```suggestion
	Handler Handler
```

This should be called Handler, otherwise you have a "stutter" "EventManager.EM" when you call it

internal/analytics/posthog/cli.go
```diff
@@ -0,0 +1,60 @@
...
+		posthog.Config{
+			Endpoint: PosthogAPIHost,
+		},
+	)
+	if err != nil {
+		return
+	}
+	defer ph.Close()
+
+	orgName := "mitb"
```
golanglemonade, line 24: Should this be hardcoded here?

internal/ent/hooks/organization.go
```diff
@@ -59,6 +61,12 @@ func HookOrganization() ent.Hook {
 				if err := createOrgMemberOwner(ctx, orgCreated.ID, mutation); err != nil {
 					return v, err
 				}
+
```
golanglemonade, line 64: this is fine for now, but might be better long term to pull this out into its own hook

internal/httpserve/middleware/auth/user.go
```diff
@@ -34,3 +34,29 @@ func GetUserIDFromContext(ctx context.Context) (string, error) {
 
 	return GetActorUserID(*ec)
 }
+
+// GetOrganizationID returns the organization ID from the echo.Context
+func GetOrganizationID(c echo.Context) (string, error) {
```
golanglemonade, line 39: Can you add a test for this, should be very similar to the existing `Test_GetActorUserID`

internal/analytics/posthog/cli.go
matoszz: yes and no - it's not used any where because I drafted it and then pivoted to implement stuff into the handlers themselves based on your feedback

PR #537, opened 2024-02-16
personal access token implementation, crud

golanglemonade:
- Adds/Updates all `CRUD` operations for personal access tokens
- Adds ability to restrict PAT to an org
- Limits the return of the actual token to only on `create` mutations
- Adds policies so users can only see their own api tokens
- Ability to set expiration, defaults to 7 days
- Updates to tokens are restricted to only the name and description

Create: 
```
go run cmd/cli/main.go pat create -n stuffff -d "for all the things" -e 730h 
{
  "createPersonalAccessToken": {
    "personalAccessToken": {
      "createdAt": "2024-02-16T16:22:05.174536-07:00",
      "createdBy": "01HPSPQP10DV3J47ZGSRH5EY1K",
      "description": "for all the things",
      "expiresAt": "2024-03-18T03:22:05.168975-06:00",
      "id": "01HPT3M8FPK40A3VVQCXZ8589Z",
      "name": "stuffff",
      "owner": {
        "displayName": "mitb@datum.net",
        "id": "01HPSPQP10DV3J47ZGSRH5EY1K"
      },
      "token": "E7XN39JOwj30ENlw26efciajleENi8m7LFVPg8ZSIehuuHG3z6fepLt3USVfF9tG",
      "updatedAt": "2024-02-16T16:22:05.174507-07:00",
      "updatedBy": "01HPSPQP10DV3J47ZGSRH5EY1K"
    }
  }
}
```

Delete: 
```
go run cmd/cli/main.go pat delete -i 01HPT3M8FPK40A3VVQCXZ8589Z
{
  "deletePersonalAccessToken": {
    "deletedID": "01HPT3M8FPK40A3VVQCXZ8589Z"
  }
}
```

Get: 
```
go run cmd/cli/main.go pat get -i 01HPSQ9CXG84WR160HSZPYS5AJ   
{
  "personalAccessToken": {
    "createdAt": "2024-02-16T12:46:26.35234-07:00",
    "createdBy": "01HPSPQP10DV3J47ZGSRH5EY1K",
    "expiresAt": "2024-02-23T12:46:26.352342-07:00",
    "id": "01HPSQ9CXG84WR160HSZPYS5AJ",
    "name": "test2",
    "owner": {
      "displayName": "mitb@datum.net",
      "id": "01HPSPQP10DV3J47ZGSRH5EY1K"
    },
    "token": "*****************************",
    "updatedAt": "2024-02-16T12:46:26.352297-07:00",
    "updatedBy": "01HPSPQP10DV3J47ZGSRH5EY1K"
  }
}
```


Taskfile.yaml
```diff
@@ -77,7 +77,13 @@ tasks:
   go:test:
     desc: runs and outputs results of created go tests
     cmds:
-      - go test -v ./...
+      - go test -v ./... 
+
+  go:test:cover:
```
matoszz, line 82: an alias for this would be nice

internal/ent/interceptors/personalaccesstoken.go
```diff
@@ -0,0 +1,38 @@
...
+	"context"
+
+	"entgo.io/ent"
+
+	"github.com/datumforge/datum/internal/ent/generated"
+	"github.com/datumforge/datum/internal/ent/generated/intercept"
+)
+
+const (
+	redacted = "*****************************"
```
matoszz, line 13: why the const / how does this work?

internal/ent/schema/personalaccesstoken.go
```diff
@@ -23,37 +25,58 @@ type PersonalAccessToken struct {
...
+			Annotations(
+				entgql.Skip(^entgql.SkipType),
+			).
 			DefaultFunc(func() string {
 				token := keygen.Secret()
 				return token
 			}),
-		field.JSON("abilities", []string{}).
-			Comment("what abilites the token should have").
+		field.String("organization_id").
```
matoszz, line 40: if the user is a part of many organizations does this require each token to be specific to 1 org? I'd think that if we did any org associations this would be `Strings` or a different relationship type

query/personalaccesstoken.graphql
```diff
@@ -19,15 +21,65 @@ mutation CreatePersonalAccessToken($input: CreatePersonalAccessTokenInput!) {
...
+      lastUsedAt
+      owner {
+        id
+        displayName
+      }
+    }
+  }
+}
+
+query GetAllPersonalAccessTokens {
```
matoszz, line 47: wouldn't the typical behavior be to search by User to see any tokens associated with them ?

internal/ent/interceptors/personalaccesstoken.go
golanglemonade: It's replacing the token with it, we can't use `Sensitive()` because the field should be returned on `Create` but should be hidden on get/update calls.

query/personalaccesstoken.graphql
golanglemonade: This returns all of the authenticated user's tokens, not all tokens in the database

matoszz (approved):

internal/ent/schema/personalaccesstoken.go
golanglemonade: updated with a join table

matoszz (approved):

PR #543, opened 2024-02-19
add sentry middlwware and helper packages

matoszz:
- inits new sentry client with our DSN and basic config wrappers
- adds sentry middleware package with skipper and default config
- configures middleware on server startup
- adds sentry utility func for sdk identifier / wrapper (needed due to echox package diff)
- adds hacky response dumper util package


internal/httpserve/serveropts/option.go
```diff
@@ -440,3 +443,23 @@ func WithSessionManager(rc *redis.Client) ServerOption {
...
+			panic(err)
+		}
+
+		s.Config.Server.Sentry = *config
+
+		// add sentry middleware
+		s.Config.Server.DefaultMiddleware = append(s.Config.Server.DefaultMiddleware, sentry.New())
+
+		// add sentry middleware to graph routes
+		s.Config.Server.GraphMiddleware = append(s.Config.Server.GraphMiddleware, sentry.New())
```
golanglemonade, line 463: DefaultMiddleware is on all the routes, you shouldnt need to specifically add it to the GraphMiddleware

internal/utils/sentry/config.go
```diff
@@ -0,0 +1,62 @@
...
+	SampleRate         float64 `split_words:"true" default:"0.2"`
+	TracesSampleRate   float64 `split_words:"true" default:"0.2"`
+	ProfilesSampleRate float64 `split_words:"true" default:"0.2"`
+	Repanic            bool    `ignored:"true"`
+	Debug              bool    `default:"false"`
+}
+
+// UseSentry true if Sentry is enabled (e.g. a DSN is configured)
+func (c Config) UseSentry() bool {
+	return c.DSN != ""
```
golanglemonade, line 28: Should this also check `c.Enabled`?

internal/httpserve/server/server.go
```diff
@@ -42,6 +43,8 @@ func NewServer(c config.Server, l *zap.SugaredLogger) *Server {
 
 // StartEchoServer creates and starts the echo server with configured middleware and handlers
 func (s *Server) StartEchoServer(ctx context.Context) error {
+	_ = sentry.Init(s.config.Sentry.ClientOptions())
```
golanglemonade, line 46: Any reason not to error check this instead of ignoring it?

internal/httpserve/serveropts/option.go
matoszz: but don't you want the ability to turn it off on both?

internal/httpserve/server/server.go
matoszz: I assumed if it passed the config checks but failed due to missing variables or otherwise the best behavior would be to fail open vs. failing the error check ?

matoszz: its a non critical, optional service for telemetry ¯\_(ツ)_/¯

internal/httpserve/serveropts/option.go
golanglemonade: that wont do it though. `DefaultMiddleware` is added to _all_ routes. So adding it to `GraphMiddleware` is duplicate

internal/httpserve/server/server.go
golanglemonade: It's non critical, but its going to silently fail. I would at least log  an error

golanglemonade (approved):

PR #593, opened 2024-02-28
Update OpenAPI Specifications

matoszz:
- consolidates generate tasks and updates `generate.go` to echo out the respective steps
- correctly sets Datum information within `entc.go` which drives the name, description, servers, etc.
- updates org owned and user owned mixins to have a OASGeneration exclusion variable
- adds ogen / oas annotations to our schemas to correct the openapi generated output (or at least get it as close to correct as we can today)
- Adds some basic errors and struct tags to the rout error handling to be re-used in later PR's


internal/ent/entc.go
```diff
@@ -45,16 +44,34 @@ func main() {
...
 		entoas.Mutations(func(graph *gen.Graph, spec *ogen.Spec) error {
-			spec.Info.SetTitle("Datum API").
+			spec.SetOpenAPI("3.1.0")
+			spec.SetServers([]ogen.Server{
+				{
+					URL:         "https://api.datum.net",
+					Description: "Datum Production API Endpoint",
+				},
+				{
+					URL:         "http://localhost:17608/v1",
```
golanglemonade, line 57: Without the v1 to align with the production server?

```suggestion
					URL:         "http://localhost:17608,
```

matoszz: prod server should actually have the v1, not the other way around

PR #609, opened 2024-03-05
Add webauthn provider

golanglemonade:
- Adds a webauthn provider
- Config settings added to example, with default settings should work for datum-ui development
- testutil added to `pkg/testutils/login` to test the flows of register + login 
- updates sessions to be of type `map[string]any` instead of `map[string]string`
- Sets default cookie settings to `Secure: true` to fix CORS issues with local development

Example Register:
image

Example Login: 
image


pkg/providers/webauthn/user.go
```diff
@@ -0,0 +1,104 @@
...
+	s, ok := Sessions[id]
+	if !ok {
+		return nil, ErrSessionNotFound
+	}
+
+	return s, nil
+}
+
+// InsertUser adds the user to the Users map
+func InsertUser(id string, user *User) {
```
matoszz, line 81: is this leftover cruft?

matoszz (approved):

Issue #643, opened 2024-03-11
port all middleware to use koanf configs


matoszz:
Today, several of the in-place middlewares are not leveraging koanf for their configurations and are missing configuration parameters entirely. Each middleware should ideally have a bool for enabled or not such that they can be toggled on or off via serverOpts and the respective inputs should also take / use koanf based configs

PR #644, opened 2024-03-11
Minor uplift to rate limiter

matoszz:
Helps address #643 by adding some of the rate limiter configurations to koanf, and toggles the rate limiter on/off via server opts. 

Still need to refactor the use of the middleware by the handlers but for now simply increasing the rate limit thresholds to avoid hitting errors when using our Taskfiles


internal/httpserve/serveropts/option.go
```diff
@@ -381,3 +381,12 @@ func WithOTP() ServerOption {
 		}
 	})
 }
+
+// WithRateLimiter sets up the rate limiter for the server
+func WithRateLimiter() ServerOption {
```
golanglemonade, line 386: This needs to be called after `WithMiddleware()` in serve.go: https://github.com/datumforge/datum/blob/main/cmd/serve.go#L54, otherwise this isn't being used at all

PR #674, opened 2024-03-19
first cut at oasrouter and wrappers

matoszz:
- adds oasrouter package with type generic handlerFunc + Route for extension / import with servers other than Echo
- adds echo specific types with the imported generics
- adds new basic /oas endpoint which creates an `echo.Route` but uses the oasrouter to wrap it and add it to the echo server - server starts up and responds on that endpoint
- some other misc updates (removal of unused functions)


matoszz:
Closing in favor of a different implementation approach

PR #689, opened 2024-03-25
use echo.bind over QueryParams and JSON decoder

golanglemonade:
Adds a `struct` for all handlers `Request` and uses `echo.Bind` instead of `ctx.QueryParams` to obtain the values. 

**Breaking Changes**: 
- `v1/invite` is now a `GET` instead of `POST`
- `v1/password-reset` now expects the `token` in the body of the request, with the password, instead of a query string param
- Strict requirement on the  `- H "Content-Type: application/json"` to be set on requests

### DIFF
--- a/internal/httpserve/handlers/resetpassword.go
+++ b/internal/httpserve/handlers/resetpassword.go
@@ -3,7 +3,6 @@ package handlers
 import (
 	"context"
 	"database/sql"
-	"encoding/json"
 	"errors"
 	"net/http"
 	"strings"
@@ -25,12 +24,7 @@ import (
 // ResetPasswordRequest contains user input required to reset a user's password
 type ResetPasswordRequest struct {
 	Password string `json:"password"`
-}
-
-// ResetPassword contains the full request to validate a password reset
-type ResetPassword struct {
-	Password string
-	Token    string
+	Token    string `json:"token"`
 }
 
 // ResetPasswordReply is the response returned from a non-successful password reset request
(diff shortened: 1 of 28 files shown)


internal/httpserve/handlers/resetpassword.go
```diff
@@ -25,12 +24,7 @@ import (
 // ResetPasswordRequest contains user input required to reset a user's password
 type ResetPasswordRequest struct {
 	Password string `json:"password"`
-}
-
-// ResetPassword contains the full request to validate a password reset
-type ResetPassword struct {
-	Password string
-	Token    string
+	Token    string `json:"token"`
```
matoszz, line 27: should this be `query:` instead of `json`?

internal/httpserve/handlers/subscribe.go
```diff
@@ -16,6 +16,12 @@ import (
 	"github.com/datumforge/datum/pkg/utils/marionette"
 )
 
+// SubscribeRequest holds the fields that should be included on a request to the `/subscribe` endpoint
+type SubscribeRequest struct {
+	Email          string `query:"email" form:"email" json:"email"`
```
matoszz, line 21: Is this to say we want to be able to receive these in each of the inputs

internal/httpserve/handlers/subscribe.go
```diff
@@ -16,6 +16,12 @@ import (
...
+	OrganizationID string `query:"organization_id" form:"organization_id" json:"organization_id"`
```
matoszz, line 22: we should mark optional fields as `required: false`

internal/httpserve/handlers/resetpassword.go
golanglemonade: Since this is a `POST` request, echo Bind will override the query with the json, so we either need to require this in the body, or bind directly with something like: 

```
err := (&DefaultBinder{}).BindQueryParams(c, &payload)
```

I opted for the former to simplify the request.

internal/httpserve/handlers/subscribe.go
golanglemonade: No, whoops, I was testing this out and forgot to remove the extras

matoszz (approved):

PR #715, opened 2024-03-31
More OAS

matoszz:
- adds taskfile
- shuffles directories under `oas`
- updates ent schema with ogen parameters


pkg/utils/oas/redoc/redocly.yaml
```diff
@@ -0,0 +1,129 @@
+theme:
+  openapi:
+<<<<<<< HEAD
```
golanglemonade, line 3: Looks like a merge conflict here

golanglemonade (approved):

PR #716, opened 2024-03-31
Additional posthog events

matoszz:
Adds additional events to several actions to increase the depth that we're creating events


internal/ent/hooks/invite.go
```diff
@@ -74,6 +77,24 @@ func HookInvite() ent.Hook {
 				m.Logger.Errorw("error sending email to user", "error", err)
 			}
 
+			orgID, _ := m.OwnerID()
+			org, _ := m.Client().Organization.Query().Where(organization.ID(orgID)).Only(ctx)
+			reqID, _ := m.RequestorID()
+			requestor, _ := m.Client().User.Query().Where(user.ID(reqID)).Only(ctx)
+			email, _ := m.Recipient()
+			role, _ := m.Role()
```
golanglemonade, line 85: nit: use the `Get` instead of `Query().Where` when you are looking up by ID. 
```suggestion
			org, _ := m.Client().Organization.Get(ctx, orgID)
			reqID, _ := m.RequestorID()
			requestor, _ := m.Client().User.Get(ctx, reqID)
			email, _ := m.Recipient()
			role, _ := m.Role()
```

internal/httpserve/handlers/register.go
```diff
@@ -26,7 +27,7 @@ type RegisterRequest struct {
 	FirstName string `json:"first_name"`
 	LastName  string `json:"last_name"`
 	Email     string `json:"email"`
-	Password  string `json:"password"`
+	Password  string `json:"password,omitempty"`
```
golanglemonade, line 30: Why `omitempty`? This register endpoint is _only_ for credentials registration, so a password should be required

internal/ent/hooks/invite.go
matoszz: what's the difference?

internal/httpserve/handlers/register.go
matoszz: I didn't realize that; l assumed we used it for other purposes beyond simply password auth

internal/ent/hooks/invite.go
golanglemonade: They hit different resolvers ,the `Get` hits the "Get" resolver, and the "Query" his the "List" resolver, just simpler to use if you are looking up by id

golanglemonade (dismissed):

golanglemonade (approved):

Issue #710, opened 2024-03-28
[Feature Request] Finish Subscriber resolvers


matoszz:
We built some initial "subscribe" endpoints to allow users to subscribe to marketing content from us. We need to finalize the implementation so that users are able to not only subscribe, but:

- We track the users who have subscribed and can report on it
- We have mechanisms of "trending analytics" around subscribed users
- Users are able to unsubscribe

PR #739, opened 2024-04-05
move to graph for auth'ed calls, add hooks

golanglemonade:
- Removes the `/subscribe` and `/unsubcribe` handlers
- Completes the resolvers for `Create` and `Delete` subscribers 
- Requires an owner_id on all subscribers
- Requires an Access Token on requests, org is now determined by the claims in the token

Partially addresses: https://github.com/datumforge/datum/issues/710


golanglemonade:
new PR incoming shortly.

PR #758, opened 2024-04-09
feat: update event payload metadata to include credential provider

matoszz:
- adds auth provider to details collected via analytics funcs in graphapi
- adds additional event emission to oauth flows given /login only uses credentials
- adds event creation for unsubscribe request which we previously weren't emitting

### DIFF
--- a/internal/httpserve/handlers/login.go
+++ b/internal/httpserve/handlers/login.go
@@ -75,7 +75,7 @@ func (h *Handler) LoginHandler(ctx echo.Context) error {
 		Set("user_id", user.ID).
 		Set("email", user.Email).
 		Set("organization_id", claims.OrgID).
-		Set("auth_provider", "CREDENTIALS")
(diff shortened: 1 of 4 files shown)


internal/httpserve/handlers/login.go
```diff
@@ -74,7 +74,8 @@ func (h *Handler) LoginHandler(ctx echo.Context) error {
 	props := ph.NewProperties().
 		Set("user_id", user.ID).
 		Set("email", user.Email).
-		Set("organization_id", claims.OrgID)
+		Set("organization_id", claims.OrgID).
+		Set("auth_provider", "CREDENTIALS")
```
golanglemonade, line 78: Can you use `user.AuthProvider` instead of hardcoding? 
```suggestion
		Set("auth_provider", user.AuthProvider)
```

golanglemonade (approved):

Issue #708, opened 2024-03-28
[Feature Request] Org switcher


matoszz:
When our customers are logged into the UI they will have an active session in 1 Organization / workspace - this will display as a drop down (the organizations they have access to, and are currently viewing) so we need to be able to:

- Allow the user to select any organization they have access to as the default they should be logged into post-authentication
- Display the current organization they are logged into
- Allow to switch organization contexts

The last bullet is the main area of code which does not exist today - we need to be able to have a "switcher" endpoint which moves the users' currently logged in organization to a different one and then re-loads the information on the page which is specific to the new organization

PR #823, opened 2024-04-19
feat: Organization Switcher

matoszz:
- Adds organization switcher to pass a target context and re-issue claims / tokens with the requested org
- adds datumclient + CLI for org switching
- adds corresponding route and uses authentication middleware

Resolves https://github.com/datumforge/datum/issues/708


cmd/cli/cmd/switch/switch.go
```diff
@@ -0,0 +1,65 @@
...
+
+var loginCmd = &cobra.Command{
+	Use:   "switch",
+	Short: "switch organization contexts",
+	RunE: func(cmd *cobra.Command, args []string) error {
+		if err := switchorg(cmd.Context()); err != nil {
+			return err
+
+		}
+		return nil
```
golanglemonade, line 23: you can simplify this 
```suggestion
		return switchorg(cmd.Context())
```

cmd/cli/cmd/switch/switch.go
```diff
@@ -0,0 +1,65 @@
...
+
+	SwitchOrganizationReply, err := datumclient.Switch(dc, ctx, input, cli.AccessToken)
+	if err != nil {
+		return err
+	}
+
+	s, err = json.Marshal(SwitchOrganizationReply)
+	if err != nil {
+		return err
+	}
```
golanglemonade, line 62: This variable doesn't need to be exported

```suggestion
	switchOrganizationReply, err := datumclient.Switch(dc, ctx, input, cli.AccessToken)
	if err != nil {
		return err
	}

	s, err = json.Marshal(switchOrganizationReply)
	if err != nil {
		return err
	}
```

internal/httpserve/handlers/switch.go
```diff
@@ -0,0 +1,138 @@
...
+	orgID, err := auth.GetOrganizationIDFromContext(context)
+	if err != nil {
+		h.Logger.Errorw("unable to get organization id from context", "error", err)
+
+		return ctx.JSON(http.StatusBadRequest, rout.ErrorResponse(err))
+	}
+
+	// ensure the user is not already in the target organization
+	if orgID == req.TargetOrganizationID {
+		return ctx.JSON(http.StatusBadRequest, rout.ErrorResponse("already switched to organization"))
```
golanglemonade, line 66: Why throw an error here? Does it matter that the user is already in it?

internal/httpserve/handlers/switch.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	// ensure user is already a member of the destination organization
+	if err := h.confirmOrgMembership(userCtx, userID, req.TargetOrganizationID); err != nil {
+		return ctx.JSON(http.StatusBadRequest, rout.ErrorResponse(err))
+	}
```
golanglemonade, line 72: Can we use FGA for this? e.g. something like this: 

```
allowed, err := h.DBClient.Authz.CheckOrgAccess(context, userID, req.TargetOrganizationID, fgax.CanView)
...
```

internal/httpserve/handlers/switch.go
```diff
@@ -0,0 +1,138 @@
...
+
+	// create new claims for the user
+	newClaims := switchClaims(user, req.TargetOrganizationID)
+
+	newClaims.UserID = user.ID
```
golanglemonade, line 77: this is already set by the `switchClaims` function above

internal/httpserve/handlers/switch.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	// return the session value for the UI to use
+	session, err := sessions.SessionToken(ctx.Request().Context())
+	if err != nil {
+		return ctx.JSON(http.StatusInternalServerError, rout.ErrorResponse(err))
+	}
+
+	// last sanity check ensuring that newly issued claims match the requested organization ID, which does not match what the user starts with, and the target organization ID which the user is already a member
+	if newClaims.OrgID != req.TargetOrganizationID {
```
golanglemonade, line 101: how would this not match since its set by the `switchClaims` function

matoszz: `claims` is a struct of `tokens` and could be modified concurrently - since we're not setting a mutex on the data modifications should you not ensure it's set correctly before returning the response?

internal/httpserve/handlers/switch.go
matoszz: we wouldn't want to issue new tokens and claims if the user already has them and there's no reason to perform things like a UI page reload if we can avoid it - is there some reason why we should respond "success" for an operation that wasn't actually successful given the purpose is claims modification?

internal/httpserve/handlers/switch.go
matoszz: is FGA more performant with the lookup? or whats the driver for the use of FGA in the REST handlers when no others use these methods?

internal/httpserve/handlers/switch.go
golanglemonade: I was thinking about the UI use case where someone switches into the org they are already in and it errors- as long as the UI handles this correctly, this makes sense to me.

internal/httpserve/handlers/switch.go
golanglemonade: FGA should generally be more performant and should be used consistently for access checks no matter if we are in a REST or graph resolver.

internal/httpserve/handlers/switch.go
golanglemonade: Yes, but the scope of the struct is limited the function so there’s no need for a mutex

internal/httpserve/handlers/switch.go
matoszz: 🤔 strange an authz store would be more performant when usually notoriously slower than an embedded database - any reason it's not used for the other checks we do in the REST handlers already then? seems like this should be sufficient but i guess i can change it if that's the preference

golanglemonade: Are we explicitly checking access to an organization anywhere else? Most of the db calls we are using are to get other information other than access checks

golanglemonade (approved):

PR #860, opened 2024-05-01
Multiple entschema additions, resolvers

matoszz:
- adds event schema and assoc edges to most major objects
- adds feature schema and entitlement relationships
- adds webhook schema and inverse edges
- adds file schema and inverse edges
- expands integration schema
- adds CLI commands for the webhook addition for now to keep PR size to a minimum


cmd/cli/cmd/webhook/webhook_create.go
```diff
@@ -0,0 +1,101 @@
...
+
+	webhookCreateCmd.Flags().StringP("url", "u", "", "the destination url the webhook is sent to")
+	datum.ViperBindFlag("webhook.create.url", webhookCreateCmd.Flags().Lookup("url"))
+
+	webhookCreateCmd.Flags().StringP("owner-id", "o", "", "owner of the webhook")
+	datum.ViperBindFlag("webhook.create.owner-id", webhookCreateCmd.Flags().Lookup("owner-id"))
+
+	webhookCreateCmd.Flags().BoolP("enabled", "e", true, "if the webhook is enabled")
+	datum.ViperBindFlag("webhook.create.enabled", webhookCreateCmd.Flags().Lookup("enabled"))
+}
```
golanglemonade, line 40: This flag isn't used in the input

internal/ent/schema/feature.go
```diff
@@ -0,0 +1,65 @@
...
+
+// Fields returns feature fields
+func (Feature) Fields() []ent.Field {
+	return []ent.Field{
+		field.String("name").
+			NotEmpty().
+			Immutable().
+			Unique(),
+		field.Bool("global").
+			Default(true),
```
golanglemonade, line 27: What does a global feature mean?

matoszz: available to everyone presumably !

internal/ent/schema/webhook.go
```diff
@@ -0,0 +1,129 @@
...
+
+// Mixin of the Webhook
+func (Webhook) Mixin() []ent.Mixin {
+	return []ent.Mixin{
+		emixin.AuditMixin{},
+		emixin.IDMixin{},
+		mixin.SoftDeleteMixin{},
+		OrgOwnerMixin{
+			Ref:        "webhooks",
+			Optional:   true,
```
golanglemonade, line 83: I don't think the org owner should be optional here

cmd/cli/cmd/webhook/webhook_create.go
```diff
@@ -68,6 +69,10 @@ func createwebhook(ctx context.Context) error {
 		DestinationURL: url,
 	}
 
+	if !enabled {
+		input.Enabled = &enabled
+	}
```
golanglemonade, line 74: This won't work always; you'll never be able to enable a webhook that was previously disabled from the cli, I used two flags to allow this to work on user settings: https://github.com/datumforge/datum/blob/main/cmd/cli/cmd/usersetting/usersetting_update.go#L43-L47

golanglemonade (approved):

Issue #866, opened 2024-05-02
[Feature Request] Claims cleanup and User Context Setting


golanglemonade:
## Authenticated User and Organization Set In Context
- All claims should contain a mapping ID to the authenticated primary key for the user id and organization id
- The primary key for the authenticated user id and organization id should be set in a context to be used for authorization checks
- This should work for user authentication as well as API key authentication, where we don't have a user id, but instead an authenticated organization and a token ID that is used in authz checks
- Personal Access tokens are allowed to be associated to multiple organizations, the context should be able to hold multiple ids and the interceptors should be updated to do a "Owner In"
- All of this should be set via the auth middleware so resolvers do not have to worry about setting any authorization context
- Context should set the authentication method, which will be required for authz checks to know how to check access

## Claims Cleanup
- JWT claims should not contain any PII (e.g. remove email address)
- Use a mapping ID instead of the database primary key

PR #870, opened 2024-05-03
revamp claims, add mapping ids squash all migrations

golanglemonade:
- Bumps entx version + adds a `mapping_id` field to every schema. This is a breaking change to the schemas since its a non-null field on existing tables so I've squashed all migrations. 
- Adds an `orgOwned` hook that will set the `owner_id` on all mutations, this eliminates the need for this to be input on the cli and the cli having to parse the claims itself. It will now create the org owned objects within the authenticated organization. 
- Updates the claims to use the user and org `mapping_id` instead of the primary key. 
- Adds the `AuthenticatedUser` context, which will contain the `subject_id`, `org_id`, and `authentication` method, added to the context during the auth middleware - this replaces the previous `ContextUserClaims` and will eventually replace all of the `viewer` context - as a single place to see the user and which organization they are authenticated into.  Since you can authenticate with an api token, the `subject_id` may be a `user_id` or the `token_id` depending on the authentication type. 
- Helper functions added `auth.GetOrganizationIDFromContext` and `auth.GetUserIDFromContext` that can be used throughout the api to determine access checks based on the user and organization. 

Example of creating a group now: 

```bash
 go run cmd/cli/main.go switch -t 01HX616NHXRXQW2EJW8H89MQ44
```

it automatically gets created within that authorized org: 

```bash
 go run cmd/cli/main.go group create -n meow             
{
  "createGroup": {
    "group": {
      "createdAt": "2024-05-05T22:08:19.104225-06:00",
      "createdBy": "01HX616C2RCX705P7NX5168VQQ",
      "description": "",
      "displayName": "meow",
      "id": "01HX61D4WM6SKJY7QEJK3F7G60",
      "logoURL": "",
      "members": [
        {
          "id": "01HX61D4X26J3B8MFR3KFNV1PG",
          "role": "ADMIN",
          "user": {
            "firstName": "matt",
            "id": "01HX616C2RCX705P7NX5168VQQ",
            "lastName": "anderson"
          }
        }
      ],
      "name": "meow",
      "owner": {
        "id": "01HX616NHXRXQW2EJW8H89MQ44"
      },

```

Relates to https://github.com/datumforge/datum/issues/866


matoszz (approved):
nice job <3 @golanglemonade

PR #920, opened 2024-05-16
Feat: httpsling

matoszz:
adds `httpsling`, a wrapper package for managing http requests and responses which has the high level capabilities:
- `RequestBuilder` constructs to easily and consistently build + send http requests
- managing cookies, middleware, TLSConfig and multiple other standard options for the `RequestBuilder`
- encoding and decoding for XML, YAML, and JSON
- default logger and ability to pass a custom logger
- middleware functions to wrap around HTTP requests, allowing pre- and post-processing of requests and responses. They can modify requests before they are sent, examine responses, and decide whether to modify them, retry the request, or take other actions
- adding / updating / managing basic auth, bearerAuth, and custom auth for requests
- Response management (status code evaluation, streaming responses, etc.)
- Retry constructs for both linear and exponential backoff


pkg/httpsling/request.go
```diff
@@ -0,0 +1,834 @@
...
+	for _, k := range key {
+		b.queries.Del(k)
+	}
+
+	return b
+}
+
+// QueriesStruct adds query parameters to the request based on a struct tagged with url tags
+func (b *RequestBuilder) QueriesStruct(queryStruct interface{}) *RequestBuilder {
+	values, _ := query.Values(queryStruct) // wafely ignore error for simplicity
```
golanglemonade, line 182: I think? 

```suggestion
	values, _ := query.Values(queryStruct) // safely ignore error for simplicity
```

pkg/httpsling/request.go
```diff
@@ -0,0 +1,834 @@
...
+			if b.client.Logger != nil {
+				b.client.Logger.Infof("Retrying request (attempt %d) after backoff", attempt+1)
+			}
+
+			// Logging context cancellation as an error condition
+			select {
+			case <-ctx.Done():
+				if b.client.Logger != nil {
+					b.client.Logger.Errorf("Request canceled or timed out: %v", ctx.Err())
+				}
```
golanglemonade, line 537: all these checks make this pretty dense, I would consider wrapping this in a func like:

```go
func (b *RequestBuilder)logError(errMsg string){
	if b.client.Logger != nil {
		b.client.Logger.Errorf(errMsg)
	}
}
```

And then here you can just have :
```go
b.logError(fmt.Sprintf("Error after %d attempts: %v", attempt+1, lastErr))
```

pkg/httpsling/response.go
```diff
@@ -0,0 +1,297 @@
...
+		Client:      client,
+	}
+
+	if response.stream != nil {
+		go response.handleStream()
+	} else {
+		if err := response.handleNonStream(); err != nil {
+			return nil, err
+		}
+	}
```
golanglemonade, line 55:
```suggestion
	if response.stream != nil {
		go response.handleStream()
	} else if err := response.handleNonStream(); err != nil {
		return nil, err
	}
```

golanglemonade (commented):
nice 👏

golanglemonade (approved):

PR #953, opened 2024-05-26
join the party fun; start a soiree pool instead of using channels

golanglemonade:
takes advantage of the new `soiree` package to handle go routines + timeouts on search. For now I've set a max search timeout of 30 seconds to return results before erroring out. This also simplies the functions to not have to use channels and lets soiree do the lifting there

### DIFF
--- a/internal/graphapi/search.resolvers.go
+++ b/internal/graphapi/search.resolvers.go
@@ -6,14 +6,9 @@ package graphapi
 
 import (
 	"context"
-	"time"
 
 	"github.com/datumforge/datum/internal/ent/generated"
 	"github.com/datumforge/datum/pkg/events/soiree"
-)
-
-var (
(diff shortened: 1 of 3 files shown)


internal/graphapi/search.resolvers.go
```diff
@@ -8,58 +8,72 @@ import (
...
 
-	groupResults := make(chan searchResult[[]*generated.Group])
-	go searchGroups(ctx, query, groupResults)
+	var (
+		orgResults []*generated.Organization
+		orgErr     error
 
-	userResults := make(chan searchResult[[]*generated.User])
-	go searchUsers(ctx, query, userResults)
+		groupResults      []*generated.Group
```
matoszz, line 23: You can pass `soiree` custom error handler via `WithErrorHandler` that might be easier to do than this setup

matoszz (approved):

PR #980, opened 2024-06-03
use local graphql hook

golanglemonade:
The existing hook was attempting to format `schema.graphl`, which is auto generated and so we would constantly have a generate diff. This changes it to only format the `schema` and `query` directory `.graphql` files.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,7 +15,11 @@ repos:
     rev: v1.20.7
     hooks:
       - id: typos
-  - repo: https://github.com/leehambley/pre-commit-graphql.git
-    rev: 97d85ddd4900ea403743d9d3ab6d8515ddbefd16
+  - repo: local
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -15,7 +15,11 @@ repos:
...
     hooks:
       - id: typos
-  - repo: https://github.com/leehambley/pre-commit-graphql.git
-    rev: 97d85ddd4900ea403743d9d3ab6d8515ddbefd16
+  - repo: local
     hooks:
       - id: format-graphql
+        name: Format GraphQL
+        description: This hook formats GraphQL schemas
+        entry: npx format-graphql@1.4.0 --write=true
```
matoszz, line 23: it feels weird to use npm / npx for this - is there no go-baed graphql formatting options?

matoszz (approved):

PR #999, opened 2024-06-07
update taskfile and precommit configuration

matoszz:
- updates goreleaser to newly changed syntax for brews
- runs auto update on existing precommit hooks
- adds a default task responder that lists available tasks
- refactor of `task install` to optimize how we approach package installations to be more comprehensive


Taskfile.yaml
```diff
@@ -154,3 +135,76 @@ tasks:
...
+    silent: true
+    desc: check if Homebrew is installed
+    cmds:
+      - '[ -x "$(command -v brew)" ] || (echo "Homebrew is not installed, please install it from https://brew.sh" && exit 1)'
+
+  go:dist:
+    desc: Downloads a Golang distribution
+    silent: true
+    vars:
+      GO_VERSION: 1.22.4
```
golanglemonade, line 205: Do we have to specify a patch? Can we tell it to just to get the latest minor so we don't have to constantly update this?

db/Taskfile.yaml
```diff
@@ -22,17 +21,6 @@ tasks:
         read migration;
         go run create_migrations.go ${migration};
 
-  lint:
```
golanglemonade, line 25: Ca we keep this command but just not call it on `task pr`?

db/Taskfile.yaml
```diff
@@ -12,7 +12,6 @@ tasks:
     aliases: [all]
     cmds:
       - task: create
-      - task: lint
```
golanglemonade, line 15: should just remove `create-and-lint` completely

Taskfile.yaml
matoszz: uh... do you know of a fancy way to do that? lol

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/datumforge/datum
 
-go 1.22.2
+go 1.22.3
+
+toolchain go1.22.4
```
golanglemonade, line 5:
```suggestion
go 1.22.4
```

I think you should just be able to do that ^ and do a `tidy`

Issue #1006, opened 2024-06-10
[Feature Request] Don't auto-set avatar, the UI will handle the default avatar icon


golanglemonade:
Right now we have a cute little robot default avatar, so the user has some avatar and its not just empty:

https://github.com/datumforge/datum/blob/main/internal/ent/hooks/user.go#L46-L47

However, the `datum-ui` has the functionality to default to an avatar with the user's initials. In order to prevent the UI from having to figure out if its the "backend" default or the user created it, we should just leave this is as `null` if its not provided.

PR #1013, opened 2024-06-11
remove avatar default

slevine12:
Removes default avatars so that the UI can handle defaults. 

resolves: https://github.com/datumforge/datum/issues/1006


internal/ent/hooks/user.go
```diff
@@ -17,7 +17,7 @@ import (
 	"github.com/datumforge/datum/internal/ent/generated/privacy"
 	"github.com/datumforge/datum/pkg/auth"
 	"github.com/datumforge/datum/pkg/passwd"
-	"github.com/datumforge/datum/pkg/utils/gravatar"
+	// "github.com/datumforge/datum/pkg/utils/gravatar"
```
matoszz, line 20:
```suggestion

```

golanglemonade (approved):

Issue #814, opened 2024-04-17
[Bug] Re-fetch google / github / social provider avatars every login


matoszz:
**Describe the bug or issue you're encountering**

If a default avatar or an avatar image is set on initial sign up, and the user changes the avatar, we do not re-fetch the image

**What are the relevant steps to reproduce, including the version(s) of the relevant software?**

Hide your google profile avatar, sign up for a datum account and view the default assigned avatar image (a robot); make your avatar public, log back into datum, the robot image is not changed to be the user's avatar

**What is the expected behavior?**

Every time I change my github or google avatar image, I see that same image when I log into Datum

PR #1084, opened 2024-07-01
refetch google/github/social provider avatars every login

slevine12:
Resolves: https://github.com/datumforge/datum/issues/814

Updated the model to have the Image in the request body. Updated the handler to pass the image through to the check and create user. Update the `createUserInput` to set the `avatar_remote_url` to the image passed in. 

Additionally, updated all of the tests to compare the remote avatar URL to the image URL.


internal/httpserve/handlers/ent.go
```diff
@@ -425,9 +432,25 @@ func createUserInput(name, email string, provider enums.AuthProvider) ent.Create
...
+	if _, err := transaction.FromContext(ctx).
+		User.UpdateOneID(entUser).
+		SetAvatarRemoteURL(image).
+		Save(ctx); err != nil {
+		h.Logger.Errorw("error updating user avatar", "error", err)
+		return err
+	}
+
+	return nil
+}
```
golanglemonade, line 452: You should add a few checks here: 

1. Check that the image is not empty: 
```
if image == "" {
    return nil 
}

2. The existing image isnt the new url, no need to update if they are equal
```
if entUser.AvatarRemoteURL != nil && *entUser.AvatarRemoteURL == image {
    return nil
}

internal/httpserve/handlers/ent_test.go
```diff
@@ -16,74 +17,100 @@ func TestCreateUserInput(t *testing.T) {
...
-				AuthProvider: &enums.AuthProviderWebauthn,
-				LastSeen:     lo.ToPtr(time.Now().UTC()),
+				FirstName:       &firstName,
+				LastName:        &lastName,
+				Email:           email,
+				AuthProvider:    &enums.AuthProviderWebauthn,
+				LastSeen:        lo.ToPtr(time.Now().UTC()),
+				AvatarRemoteURL: &image,
 			},
+			image: image,
```
golanglemonade, line 87: webauthn provider would never have an image, so this should check with an empty string
```suggestion
			image: image,
```

golanglemonade (approved):

Issue #1059, opened 2024-06-25
[Bug] Token `LastUsed` not updated


golanglemonade:
**Describe the bug or issue you're encountering**

When using a personal_access_token and ap_tokens, the `lastUsed` field is not being updated

**What are the relevant steps to reproduce, including the version(s) of the relevant software?**

Create a PAT
```
✗ go run cmd/cli/main.go pat create -n stuffff -d "for all the things" -e 730h -o 01J18E87D587GKK8K18MPYWER4
```

Check `last_used`

```bash
go run cmd/cli/main.go pat get                                                                            
  ID                          NAME     TOKEN                          LASTUSEDAT  EXPIRESAT                             
  01J18GR4DJH5Q17QA5J6J4C9E1  stuffff  *****************************  never       2024-07-26 06:17:43.84478 +0000 UTC   
  01J18GTHP3E244W8NHBGMZYTNF  stuffff  *****************************  never       2024-07-26 06:19:02.960658 +0000 UTC  
```

Use the PAT in a request, and recheck `last_used` - it will still be `never`

**What is the expected behavior?**

`LastUsedAt` is updated when PAT and API tokens are used

**Code Location**

This should be updated in the auth middleware that determines if the Token is valid before the `AuthenticatedUser` is returned https://github.com/datumforge/datum/blob/main/pkg/middleware/auth/auth.go

PR #1149, opened 2024-07-22
Bug token lastused

slevine12:
Resolves: https://github.com/datumforge/datum/issues/1059

The `LastUsedAt` field is now updated when PAT and API tokens are used. The function `func updateLastUsed(ctx context.Context, dbClient *generated.Client, au *auth.AuthenticatedUser, tokenID string)` was added in the `auth.go` file to successfully update the `LastUsedAt` field. This function was referenced in `func Authenticate(conf *AuthOptions) echo.MiddlewareFunc` (also found in the `auth.go` file) in order to update this field while tokens are being authenticated. 

Tests were also implemented in both the `personalaccesstoken_test.go` and `apitoken_test.go` files to make sure `LastUsedAt` was being updated correctly.


internal/graphapi/apitoken_test.go
```diff
@@ -368,3 +370,38 @@ func (suite *GraphTestSuite) TestMutationDeleteAPIToken() {
 		})
 	}
 }
+
+func (suite *GraphTestSuite) TestLastUsedAPIToken() {
+
```
golanglemonade, line 375:
```suggestion
```

internal/graphapi/apitoken_test.go
```diff
@@ -368,3 +370,38 @@ func (suite *GraphTestSuite) TestMutationDeleteAPIToken() {
...
+	t := suite.T()
+
+	defer mock_fga.ClearMocks(suite.client.fga)
+
+	// setup user context
+	reqCtx, err := userContext()
+	require.NoError(t, err)
+
+	// create new API token
+	token := (&APITokenTokenBuilder{client: suite.client}).MustNew(reqCtx, t)
```
golanglemonade, line 385: This naming was fixed between your changes, so you need to update it to match.
```suggestion
	token := (&APITokenBuilder{client: suite.client}).MustNew(reqCtx, t)
```

internal/graphapi/apitoken_test.go
```diff
@@ -368,3 +370,38 @@ func (suite *GraphTestSuite) TestMutationDeleteAPIToken() {
...
+
+	mock_fga.CheckAny(t, suite.client.fga, true)
+
+	// check that the last used is empty
+	res, err := suite.client.datum.GetAPITokenByID(reqCtx, token.ID)
+	require.NoError(t, err)
+	assert.Empty(t, res.APIToken.LastUsedAt)
+
+	// TODO: (slevine: update once we have updated the last used at field on the token when used)
+	// // setup graph client using the API token
```
golanglemonade, line 395: remove old comment
```suggestion
	// setup graph client using the API token
```

internal/graphapi/personalaccesstoken_test.go
```diff
@@ -426,15 +428,15 @@ func (suite *GraphTestSuite) TestLastUsedPersonalAccessToken() {
 
 	// TODO: (slevine: update once we have updated the last used at field on the token when used)
 	// // setup graph client using the personal access token
```
golanglemonade, line 430:
```suggestion
	// setup graph client using the personal access token
```

pkg/middleware/auth/auth.go
```diff
@@ -65,14 +65,16 @@ func Authenticate(conf *AuthOptions) echo.MiddlewareFunc {
 
 			var au *auth.AuthenticatedUser
 
+			var id string
+
```
golanglemonade, line 69: this is a nit, but if you are declaring multiple variables in a row its usually better to do: 

```suggestion
			var  (
				au *auth.AuthenticatedUser
				id string
			)
```

golanglemonade (approved):

matoszz (approved):

PR #1182, opened 2024-08-03
Remove dead / unused code

matoszz:
- removes sentry package and middleware (will be implemented at a later date as a customer-facing configuration option)
- removes unused pdf package
- removes storage package in preparation for a replacement that will come in a separate PR
- removes unused analytics functions and some associated calls
- removes unused jsonschema models


cmd/root.go
```diff
@@ -47,19 +45,12 @@ func init() {
...
 	if err := initCmdFlags(rootCmd); err != nil {
 		log.Fatalf("error loading config: %v", err)
 	}
-
-	c := sentry.LoggerConfig{
-		Debug:  k.Bool("debug"),
-		Pretty: k.Bool("pretty"),
-	}
-
-	logger = c.NewLogger()
```
golanglemonade, line 60: we still need to setup the logger, otherwise the server won't startup because of a nil pointer error, 

this was ported over to the sentry package but previously was: 

```
func setupLogging() {
	cfg := zap.NewProductionConfig()
	if viper.GetBool("pretty") {
		cfg = zap.NewDevelopmentConfig()
	}

	if viper.GetBool("debug") {
		cfg.Level = zap.NewAtomicLevelAt(zap.DebugLevel)
	} else {
		cfg.Level = zap.NewAtomicLevelAt(zap.InfoLevel)
	}

	l, err := cfg.Build()
	if err != nil {
		panic(err)
	}

	logger = l.Sugar().With("app", appName)
	defer logger.Sync() //nolint:errcheck
}
```

cmd/root.go
```diff
@@ -47,19 +46,35 @@ func init() {
...
 	return k.Load(posflag.Provider(cmd.Flags(), k.Delim(), k), nil)
 }
+
+func setupLogging() {
+	cfg := zap.NewProductionConfig()
+	if viper.GetBool("pretty") {
+		cfg = zap.NewDevelopmentConfig()
+	}
+
+	if viper.GetBool("debug") {
```
golanglemonade, line 67: This should no longer be viper, and instead the koanf config here and line 63 (and remove the viper import that got added)

```suggestion
	if k.Bool("debug") {
```

cmd/root.go
```diff
@@ -47,19 +46,35 @@ func init() {
...
 }
 
+// initCmdFlags loads the flags from the command line into the koanf instance
 func initCmdFlags(cmd *cobra.Command) error {
 	return k.Load(posflag.Provider(cmd.Flags(), k.Delim(), k), nil)
 }
+
+func setupLogging() {
+	cfg := zap.NewProductionConfig()
+	if viper.GetBool("pretty") {
```
golanglemonade, line 63: This one is still here, I didn’t do a full suggested comment because I also need to remove the viper import

golanglemonade, line 63:
```suggestion
	if k.Bool("pretty") {
```

cmd/root.go
```diff
@@ -6,9 +6,8 @@ import (
 	"github.com/knadh/koanf/providers/posflag"
 	"github.com/knadh/koanf/v2"
 	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
```
golanglemonade, line 9:
```suggestion
```

golanglemonade (approved):

PR #1200, opened 2024-08-07
feat: check access handler

golanglemonade:
Adds a check access handler that will query the FGA API for the current authrozied user and their access to an object with a given relation:

```
curl -s -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" http://localhost:17608/v1/check-access -d '{"objectId":"01J4HMNDSZCCQBTY93BF9CBF5D", "objectType":"organization", "relation":"member"}'  | jq
{
  "success": true,
  "allowed": true
}
```


internal/httpserve/handlers/checkaccess.go
```diff
@@ -0,0 +1,69 @@
...
+	checkAccess := openapi3.NewOperation()
+	checkAccess.Description = "Check Subject Access to Object"
+	checkAccess.OperationID = "CheckAccess"
+	checkAccess.Security = &openapi3.SecurityRequirements{}
+
+	h.AddRequestBody("CheckAccessRequest", models.ExampleInviteRequest, checkAccess)
+	h.AddResponse("CheckAccessReply", "success", models.ExampleInviteResponse, checkAccess, http.StatusOK)
+	checkAccess.AddResponse(http.StatusInternalServerError, internalServerError())
+	checkAccess.AddResponse(http.StatusBadRequest, badRequest())
+	checkAccess.AddResponse(http.StatusUnauthorized, unauthorized())
```
matoszz, line 66: the responses in your handler are InvalidInput, BadRequest, and InternalServerError so these should match (update unauthorized to used invalidinput instead)

internal/httpserve/route/checkaccess.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	"net/http"
+
+	echo "github.com/datumforge/echox"
+)
+
+// registerCheckAccessHandler registers the check-access handler
+func registerCheckAccessHandler(router *Router) (err error) {
+	path := "/check-access"
```
matoszz, line 11: kind of just a nit, but I think this should be `/account` or similar; most REST API's (and their published openapi specs) follow pretty generally standard naming conventions and things associated with metadata about an account (including what permissions they might have) are usually under an `/account` or `/user` path so i'd personally use `/account` as we may add onto `/account` later for more details beyond just the RBAC

pkg/models/models.go
```diff
@@ -575,3 +575,57 @@ type OauthTokenRequest struct {
...
+// CHECK-ACCESS
+// =========
+
+// CheckAccessRequest holds the fields that should be included on a request to the `/check-access` endpoint
+type CheckAccessRequest struct {
+	ObjectID    string `json:"objectId"`
+	ObjectType  string `json:"objectType"`
+	SubjectID   string `json:"subjectId"`
+	SubjectType string `json:"subjectType"`
+	Relation    string `json:"relation"`
```
matoszz, line 589: are any of these possibly optional? e.g. could you pass in just the object ID and get back `can_view` or other details around it? might future proof us from needing to refactor this a bit if we passed simply the object ID + subject ID and then got the list of what they could do vs. constraining the query to include the `relation`

internal/httpserve/handlers/checkaccess.go
golanglemonade: @matoszz I'll add Invalid input, but I Think unauthorized should stay because the endpoint requires authorization so you could get an unauthorized response

pkg/models/models.go
golanglemonade: I was planning on making a different endpoint for `list objects`, which would give all the users objects for a given relation. You currently cannot get all relations from fga for a user.

internal/httpserve/handlers/checkaccess.go
golanglemonade: InvalidInput == BadRequest as far as I can tell

internal/httpserve/route/checkaccess.go
golanglemonade: Updated to `account/access`

matoszz (approved):

